PR #12, opened 2021-03-19
Message docking componentized protocol

Counterflowwind:
#### What type of PR is this?
feature
Interface docking componentized protocol

Which issue(s) this PR fixes:
Fixes #


bundle/notify.go
```diff
@@ -0,0 +1,156 @@
+package bundle
+
+import (
+	"fmt"
+	"github.com/erda-project/erda/apistructs"
```
Effet, line 5: separate by empty line

bundle/notify_test.go
```diff
@@ -0,0 +1,37 @@
+package bundle
+
+import (
+	"fmt"
```
Effet, line 4: bad format

Effet (changes requested):

Effet (approved):

PR #19, opened 2021-03-22
feature: add pipeline and actionagent

sfwn:
#### What type of PR is this?

feature

add pipeline & actionagent to erda

### DIFF
--- a/modules/actionagent/agenttool/tar.go
+++ b/modules/actionagent/agenttool/tar.go
@@ -1,0 +1,95 @@
+package agenttool
+
+import (
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/mholt/archiver"
+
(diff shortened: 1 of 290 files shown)


modules/actionagent/agenttool/tar.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/erda-project/erda/pkg/retry"
+
+	"github.com/mholt/archiver"
```
Effet, line 12: bad order

Effet (changes requested):

modules/actionagent/agenttool/tar.go
sfwn: adjusted.

Effet (approved):
a horrible huge PR!

PR #24, opened 2021-03-22
feature: add healthcheck pkg

sfwn:
#### What type of PR is this?

feature

add healthcheck pkg

### DIFF
--- a/pkg/healthcheck/monitor.go
+++ b/pkg/healthcheck/monitor.go
@@ -1,0 +1,108 @@
+package healthcheck
+
+import (
+	"context"
+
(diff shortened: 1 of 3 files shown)


liuhaoyang (changes requested):
What is the difference with [infra-providers/health](https://github.com/erda-project/erda-infra/blob/develop/providers/health)

sfwn:
> What is the difference with [infra-providers/health](https://github.com/erda-project/erda-infra/blob/develop/providers/health)

@kakj-go will refactor it in 4.1

PR #81, opened 2021-03-30
fix golang-cilint run

QvodSoldier:
#### What type of PR is this?
/kind bug

fix cilint run error:  context loading failed: no go files to analyze

https://github.com/golangci/golangci-lint-action/issues/23

https://github.com/argoproj/argo-rollouts/pull/579/commits/1acf99b59678840be1e72b7151e4c8fc97407cad#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9f

#### Specified Reivewers:
@Effet @sfwn

### DIFF
--- a/.github/workflows/ci-it.yml
+++ b/.github/workflows/ci-it.yml
@@ -27,15 +27,15 @@ jobs:
           # Optional: working directory, useful for monorepos
           # working-directory: somedir
           # Optional: golangci-lint command line arguments.
-          # args: --issues-exit-code=0
+          # args: -v --timeout=10m
           # Optional: show only new issues if it's a pull request. The default value is `false`.
           # only-new-issues: true
           # Optional: if set to true then the action will use pre-installed Go.
           # skip-go-installation: true
           # Optional: if set to true then the action don't cache or restore ~/go/pkg.
-          # skip-pkg-cache: true
+          skip-pkg-cache: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-it.yml
```diff
@@ -9,16 +9,12 @@ on:
       - release/*
   pull_request:
 jobs:
-  CI:
+  lint-go:
+    name: Lint Go code
```
liuhaoyang, line 13: Please rename `Lint Go code` to `CI-Lint`

liuhaoyang (changes requested):

liuhaoyang (commented):
I think we only need to set skip-pkg-cache and skip-build-cache to true in the golangci-lint step, there is no need to split lint and build into 2 jobs.

QvodSoldier:
> I think we only need to set skip-pkg-cache and skip-build-cache to true in the golangci-lint step, there is no need to split lint and build into 2 jobs.

QvodSoldier:
> I think we only need to set skip-pkg-cache and skip-build-cache to true in the golangci-lint step, there is no need to split lint and build into 2 jobs.

yes, but still execute 'go mod tidy' before 'golang-cilint'

liuhaoyang (approved):

PR #88, opened 2021-03-31
fix(scheduler): update runtime label to support global service in the namespace

Muzry:
#### What type of PR is this?
/kind bug
- fix(scheduler): update runtime label to support global service in the namespace

Fixes #

#### Specified Reivewers:
/assgin @luobily @sixther-dc 

#### Does this PR introduce a user-facing change?

```release-note

```

```docs

```

### DIFF
--- a/modules/scheduler/executor/plugins/k8s/namespace/namespace.go
+++ b/modules/scheduler/executor/plugins/k8s/namespace/namespace.go
@@ -71,7 +71,7 @@ func (n *Namespace) Create(ns string, labels map[string]string) error {
 	if !resp.IsOK() {
 		return errors.Errorf("failed to create namespace, ns: %s, statuscode: %v, body: %v", ns, resp.StatusCode(), b.String())
 	}
-	logrus.Infof("succeed to create namespace, ns: %s", ns)
+	logrus.Infof("succeed to create namespace %s", ns)
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


modules/scheduler/executor/plugins/k8s/namespace/namespace.go
```diff
@@ -71,7 +71,7 @@ func (n *Namespace) Create(ns string, labels map[string]string) error {
 	if !resp.IsOK() {
 		return errors.Errorf("failed to create namespace, ns: %s, statuscode: %v, body: %v", ns, resp.StatusCode(), b.String())
 	}
-	logrus.Infof("succeed to create namespace, ns: %s", ns)
+	logrus.Infof("succeed to create namespace, ns %s from module %s", ns, module)
```
luobily, line 74: parameters should not only be used to print logs.
pls delete this param (module).

luobily (changes requested):

modules/scheduler/executor/plugins/k8s/namespace/namespace.go
Muzry: already  modify to 
```suggestion
logrus.Infof("succeed to create namespace, ns: %s", ns)
```

luobily (approved):
/lgtm

PR #94, opened 2021-04-06
add '-tag musl' build parameter to resolve confluent-kafka-go compatibility problem

sixther-dc:
#### What type of PR is this?
/kind bug

after we upgrade the version of confluent-kafka-go, it will cause the following compilation errors when building scheduler
image

It seems like it is linking the glibc librdkafka (which references __strdup).
build  with -tags musl can use the built-in librdkafka。

https://github.com/confluentinc/confluent-kafka-go/issues/454

Fixes #

#### Specified Reivewers:
/assgin @luobily @recallsong


luobily (approved):
/lgtm

luobily (approved):
/lgtm

PR #103, opened 2021-04-07
add example proto, update Makefile to check go codes

recallsong:
#### What type of PR is this?

/kind feature

1、add example proto
2、update Makefile to check go codes

#### Specified Reivewers:
@liuhaoyang 

```sh
# test、check and format all go files
make normalize
#  test、check and format all go files in MODULE_PATH
make normalize MODULE_PATH=modules/examples/example-proto 
```


Makefile
```diff
@@ -95,6 +95,14 @@ run-g: build
 run-ps: build
 	./bin/${APP_NAME} --providers
 
+# normalize all go files before push to git repo
+normalize:
+	go mod tidy
+	go test -test.timeout=10s ./...
+	find ./ -name "*.go" | xargs gofmt -w -l
+	find ./ -name "*.go" | goimports -w l
+	golint -set_exit_status ./...
```
erenming: why no golangci-lint enable? where ci action use to check

recallsong: use it at local before push git

liuhaoyang (approved):

PR #109, opened 2021-04-07
adjust the k8s dependencies in go.mod

johnlanni:
#### What type of PR is this?
/kind cleanup

1.  make the k8s version (1.18.3) of all component consistent 
2. remove useless kubelet event

#### Specified Reivewers:
/assgin @luobily luo @liuhaoyang liu


modules/scheduler/executor/plugins/k8s/event/event.go
```diff
@@ -140,7 +140,6 @@ func (e *Event) WatchPodEventsAllNamespaces(ctx context.Context, callback func(*
 			body.Close()
 		}
 	}
-	panic("unreachable")
```
luobily, line 143: Panic is still necessary to catch exceptions in the watch process

luobily (approved):
/lgtm

PR #123, opened 2021-04-08
Migrate basic warehouse 

luoyuctl:
What type of PR is this?
Migrate basic warehouse

/kind feature

For the subsequent migration of other modules

Specified Reivewers:
/assgin @liuhaoyang

Does this PR introduce a user-facing change?
NONE


pkg/docker/client.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2021 Terminus, Inc.
+
```
liuhaoyang, line 2: The `docker-client` provider should be provided by erda-infra

pkg/stream/writer/buffer.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2021 Terminus, Inc.
```
liuhaoyang, line 1: The buffer pkg already exists in erda-infra

liuhaoyang (changes requested):

liuhaoyang (changes requested):
Why are the common functions of file and json scattered in different packages such as `pkg/utils` and `module/monitor/utils` ?

luoyuctl:
> Why are the common functions of file and json scattered in different packages such as `pkg/utils` and `module/monitor/utils` ?
pkg/utils are allready deleted now

pkg/docker/client.go
luoyuctl: done

luoyuctl:
> Why are the common functions of file and json scattered in different packages such as `pkg/utils` and `module/monitor/utils` ?

luoyuctl:
> Why are the common functions of file and json scattered in different packages such as `pkg/utils` and `module/monitor/utils` ?
Duplicate of #

liuhaoyang (approved):

PR #130, opened 2021-04-09
move admin tools to erda

Counterflowwind:
What type of PR is this?
move admin-tools to erda

/kind feature

move admin-tools to erda

Specified Reivewers:
/assgin @liuhaoyang


bundle/urls.go
```diff
@@ -90,7 +90,8 @@ func (u urls) Collector() (string, error) {
 }
 
 func (u urls) Monitor() (string, error) {
-	return u.getURL(discover.EnvMonitor, discover.SvcMonitor)
+	//return u.getURL(discover.EnvMonitor, discover.SvcMonitor)
```
liuhaoyang, line 93: Don't use hard-coded config.

conf/monitor/monitor.yml
```diff
@@ -0,0 +1,29 @@
+http-server:
```
liuhaoyang, line 1: The [conf/monitor/monitor.yaml](https://github.com/erda-project/erda/blob/develop/conf/monitor/monitor.yaml) already exists, and the two config files need to be merged.

liuhaoyang (changes requested):

liuhaoyang (approved):

PR #178, opened 2021-04-13
modify set Cookie's Domain

QvodSoldier:
#### What type of PR is this?
/kind bug

In the login process, we need set cookie to subdomain(e.g. .app.xxx.com) rather than .xxx.com

#### Specified Reivewers:
/assgin @Effet


Effet (approved):

modules/openapi/conf/conf.go
```diff
@@ -172,7 +172,7 @@ func GetDomain(host, confDomain string) (string, error) {
 	confDomains := strings.SplitN(confDomain, ",", -1)
 	for _, v := range confDomains {
 		if strings.Contains(v, domain) {
-			return domain, nil
+			return v, nil
```
sfwn, line 175: please add unit test for this function.

sfwn (changes requested):

modules/openapi/conf/conf.go
QvodSoldier: it's has existed in config_test.go

sfwn (approved):

PR #186, opened 2021-04-14
speed up local build

sfwn:
#### What type of PR is this?

optimization

speed up local build


Makefile
```diff
@@ -78,17 +78,22 @@ build-version:
 	@echo ------------ End   Build Version Details ------------
 
 tidy:
-	cd "${BUILD_PATH}" && \
-    ${GO_BUILD_ENV} go mod tidy
+	if [[ -f "${BUILD_PATH}/go.mod" ]]; then \
```
recallsong, line 81: use @ to hide these commands

recallsong (changes requested):

recallsong (approved):

PR #192, opened 2021-04-14
adjust build scripts for building all-in-one image

johnlanni:
#### What type of PR is this?
/kind feature

we can use the all-in-one image in erda's self CICD pipeline.


build/dockerfiles/Dockerfile-all
```diff
@@ -0,0 +1,28 @@
...
+RUN mkdir -p "$GOPATH/src/github.com/erda-project/erda/"
+COPY . "$GOPATH/src/github.com/erda-project/erda/"
+WORKDIR "$GOPATH/src/github.com/erda-project/erda/"
+
+RUN make build-all
+
+ARG BASE_DOCKER_IMAGE
+FROM ${BASE_DOCKER_IMAGE}
+
+RUN apk add --no-cache jq
```
innerpeacez, line 13: why install jq?

johnlanni: let @recallsong explain it

Effet: there's a temporary solution:
aim to boost speed of api-test, we make api-test runs in same process as the pipeline.

so jq called by pipeline directly...

and we are proposing the final solution, will be fixed in the next iteration.

luoyuctl: @recallsong

innerpeacez: > there's a temporary solution:
> aim to boost speed of api-test, we make api-test runs in same process as the pipeline.
> 
> we use jq to achieve some function like value-extracting, assertion... so jq called by pipeline directly...
> 
> and we are proposing the final solution, will be fixed in the next iteration.

got it

recallsong (approved):

PR #197, opened 2021-04-15
cron task to clean up the repository

kakj-go:
#### What type of PR is this?
feature

If you do not clean up the git warehouse regularly, too long a time will cause the access to the warehouse to time out


modules/gittar/pkg/gc/gc.go
```diff
@@ -0,0 +1,112 @@
...
+	wait.Wait()
+}
+
+// execute the git gc command and print the returned information or error
+func doGcCommand(path string) {
+	// remove spaces to prevent injection attacks
+	path = strings.Replace(path, " ", "", -1)
+	cmd := exec.Command("sh", "-c", "cd '"+path+"' && git gc")
+	logrus.Infof("gc: start gc path: %v", path)
+	stdout, err := cmd.StdoutPipe()
```
sfwn, line 92: how to handle stderr?

modules/gittar/pkg/gc/gc.go
```diff
@@ -0,0 +1,112 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/erda-project/erda/modules/gittar/pkg/limit_sync_group"
+	"github.com/erda-project/erda/pkg/cron"
+	"github.com/sirupsen/logrus"
+)
+
+// all gittar repository root address
+const repositoryRootAddr = "/repository"
```
sfwn, line 28: Get repositoryRootAddress from config.
If someone changes the root addr one day, it is easy for him to be unclear about the gc configuration, leading to a mismatch.

modules/gittar/pkg/limit_sync_group/group.go
```diff
@@ -0,0 +1,43 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+package limit_sync_group
+
+import "sync"
+
+// encapsulate sync.WaitGroup
+// realize a controllable number of WaitGroup
+type limitSyncGroup struct {
+	c  chan struct{}
```
sfwn, line 21: use buffered channel is enough.

sfwn (changes requested):

modules/gittar/pkg/gc/gc.go
```diff
@@ -0,0 +1,112 @@
...
+
+// gc repository concurrent number
+const concurrentNum = 1
+
+// gc repository cron time
+const cronExpression = "0 0 1 * * ?"
```
QvodSoldier, line 34: do we need get cron expresssion from config? Maybe there were some scenes need different gc cycle

QvodSoldier:
I found git gc have a --auto options [git gc](https://git-scm.com/docs/git-gc), but i don't sure is right for us, i think we can use this fix for now.

PR #209, opened 2021-04-17
I saved gittar's life

QvodSoldier:
#### What type of PR is this?
/kind bug

our dockerfile has changed, gittar's workdir is not in the "/", but volume is still mount on "/"

#### Specified Reivewers:
/assgin @Effet @sfwn


modules/gittar/conf/conf.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // Conf 定义配置对象.
 type Conf struct {
-	RepoRoot        string `env:"GITTAR_REPOSITORY_ROOT" default:"./repository"`
+	RepoRoot        string `env:"GITTAR_REPOSITORY_ROOT" default:"/repository"`
```
sfwn, line 25: check bind in version's dice.yml

QvodSoldier: image

yeah, it was mounted on the root path

sfwn (approved):

PR #231, opened 2021-04-22
autotest netportal set cookie to save login

kakj-go:
#### What type of PR is this?
feature

add netportal cookie will not in request host domain the login will lost


pkg/apitestsv2/apitest.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/erda-project/erda/apistructs"
+	"github.com/erda-project/erda/modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/cookiejar"
```
sfwn, line 27: bad import.

pkg/apitestsv2/netportal.go
```diff
@@ -18,18 +18,34 @@ import (
 	"net/http"
 
 	"github.com/erda-project/erda/apistructs"
+	"github.com/erda-project/erda/modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/cookiejar"
```
sfwn, line 21: bad import.

sfwn (changes requested):

pkg/apitestsv2/netportal.go
luoyuctl: why indicate the same issue?

sfwn (approved):

PR #249, opened 2021-04-26
Fix: Request error rate calculation error

innerpeacez:
- [x]  fix request error rate calculation error
- [x]  i18n


liuhaoyang (approved):
/lgtm

PR #262, opened 2021-04-27
Make org auth compatible with multiple root domain

shuofan:
#### What type of PR is this?
feature

Make org auth compatible with multiple root domain

Fixes #

#### Specified Reivewers:
@Effet 

#### Does this PR introduce a user-facing change?

```release-note

```

```docs

```

### DIFF
--- a/modules/cmdb/services/org/org.go
+++ b/modules/cmdb/services/org/org.go
@@ -494,17 +494,19 @@ func (o *Org) GetOrgByDomain(domain string) (*model.Org, error) {
 	if domain != "" && conf.OrgWhiteList[domain] {
 		return nil, nil
 	}
-	suf := strutil.Concat(".", conf.RootDomain())
-	domain_and_port := strutil.Split(domain, ":", true)
-	domain = domain_and_port[0]
-	if !strutil.HasSuffixes(domain, suf) {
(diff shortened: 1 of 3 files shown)


modules/cmdb/services/org/org.go
```diff
@@ -494,17 +494,19 @@ func (o *Org) GetOrgByDomain(domain string) (*model.Org, error) {
...
-	suf := strutil.Concat(".", conf.RootDomain())
-	domain_and_port := strutil.Split(domain, ":", true)
-	domain = domain_and_port[0]
-	if !strutil.HasSuffixes(domain, suf) {
-		return nil, apierrors.ErrGetOrg.NotFound()
-	}
-	orgName := strutil.TrimSuffixes(domain, suf)
-	if strutil.HasSuffixes(orgName, "-org") {
-		orgName = strutil.TrimSuffixes(orgName, "-org")
+	for _, rootDomain := range conf.RootDomainList() {
```
sfwn, line 497: Split to a new method and add unit test please.

sfwn (changes requested):

modules/cmdb/services/org/org.go
shuofan: done

sfwn (approved):

PR #263, opened 2021-04-27
Task prepare step return error correctly

sfwn:
#### What type of PR is this?

bug

Some error was ate at task prepare step.


luoyuctl:
If the problem isn't urgent , you can put optimization and more bug fixes together.

Effet (approved):

PR #265, opened 2021-04-27
cron task to clean up the repository

kakj-go:
#### What type of PR is this?

feature

If you do not clean up the git warehouse regularly, too long a time will cause the access to the warehouse to time out

### DIFF
--- a/modules/gittar/pkg/gc/gc.go
+++ b/modules/gittar/pkg/gc/gc.go
@@ -1,0 +1,105 @@
+// Copyright (c) 2021 Terminus, Inc.
+//
+// This program is free software: you can use, redistribute, and/or modify
+// it under the terms of the GNU Affero General Public License, version 3
+// or later ("AGPL"), as published by the Free Software Foundation.
+//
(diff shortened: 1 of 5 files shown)


modules/gittar/pkg/gc/gc.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+
+	"github.com/erda-project/erda/modules/gittar/conf"
+	"github.com/erda-project/erda/modules/gittar/pkg/limit_sync_group"
+	"github.com/erda-project/erda/pkg/cron"
+	"github.com/sirupsen/logrus"
+)
+
+// all gittar repository root address
+var repositoryRootAddr = "/repository"
```
sfwn, line 29: use conf.RepoRoot is better ?

sfwn (changes requested):

modules/gittar/pkg/gc/gc.go
sfwn: ok i saw in line 45.

sfwn (approved):

PR #266, opened 2021-04-28
The scheduler doesn't delete namespace when set retainNamespace is true

Muzry:
#### What type of PR is this?
/kind bug

feat(prefix): change dice prefix to erda

Fixes #

#### Specified Reivewers:
/assgin @luobily


luobily (dismissed):
lgtm

luobily (approved):

PR #269, opened 2021-04-28
bug fix snippetConfig name or source same like will lost yml

kakj-go:
#### What type of PR is this?
bugfix

bug fix snippetConfig name or source same like will lost yml


modules/pipeline/services/pipelinesvc/snippet.go
```diff
@@ -15,91 +15,63 @@ package pipelinesvc
...
-			yml, err := s.queryPipelineYAMLBySnippetConfig(&sc)
-			if err != nil {
-				return nil, err
-			}
-			if _, ok := yamlResults[sc.Source]; !ok {
-				yamlResults[sc.Source] = make(map[string]string)
-			}
-			yamlResults[sc.Source][sc.Name] = yml
+	for _, singleConfig := range singleSourceSnippetConfigs {
+		yml, err := s.queryPipelineYAMLBySnippetConfig(&singleConfig)
```
sfwn, line 56: async handle

modules/pipeline/services/pipelinesvc/snippet.go
```diff
@@ -15,91 +15,63 @@ package pipelinesvc
...
+		if err != nil {
+			return nil, err
 		}
+		yamlResults[jsonparse.JsonOneLine(singleConfig)] = yml
+	}
+
+	var errMsgs []string
+	for _, snippetConfig := range sourceSnippetConfigs {
+		yml, ok := yamlResults[jsonparse.JsonOneLine(snippetConfig)]
```
sfwn, line 65: please test whether jsonparse.JsonOneLine will print the same result while config has map field.

if not same, please change map to character-ordered string slice and then json marshal.

modules/pipeline/services/pipelinesvc/create.go
```diff
@@ -330,20 +331,17 @@ func (s *PipelineSvc) createPipelineGraph(p *spec.Pipeline, passedDataOpt ...pas
 
 	// 统一处理嵌套流水线
 	// 批量查询 snippet yaml
-	sourceSnippetConfigMap := make(map[string]map[string]apistructs.SnippetConfig) // key: source, value: type
+	var sourceSnippetConfig []apistructs.SnippetConfig
```
sfwn, line 334: configs

sfwn (changes requested):

modules/pipeline/services/pipelinesvc/snippet.go
kakj-go: Whether the subsequent pipeline.yml query of the snippet provider should be a batch query interface, so there is no need to async query here, if async query gittar,query pressure will be too high

apistructs/pipeline_yml.go
```diff
@@ -111,6 +118,36 @@ type SnippetConfig struct {
...
+		for k, v := range snippetConfig.Labels {
+			snippetLabels = append(snippetLabels, SnippetLabel{
+				Key:   k,
+				Value: v,
+			})
+		}
+		sort.Sort(snippetLabels)
+		labels = jsonparse.JsonOneLine(snippetLabels)
+	}
+	return fmt.Sprintf("{\"source\": %s, \"name\": %s, \"labels\": %s}", snippetConfig.Source, snippetConfig.Name, labels)
```
sfwn, line 148: Values of labels will have quote problems.
You can define a anonymous struct and json marshal it.

modules/pipeline/services/pipelinesvc/snippet.go
```diff
@@ -25,81 +26,51 @@ import (
...
-		for _, sc := range singleConfigs {
-			yml, err := s.queryPipelineYAMLBySnippetConfig(&sc)
-			if err != nil {
-				return nil, err
-			}
-			if _, ok := yamlResults[sc.Source]; !ok {
-				yamlResults[sc.Source] = make(map[string]string)
-			}
-			yamlResults[sc.Source][sc.Name] = yml
+	for _, singleConfig := range singleSourceSnippetConfigs {
```
sfwn, line 54: You can handle batch & single at the same time.

sfwn (changes requested):

sfwn:
Please structure a scenario and test time costed before and after this pr.

sfwn (approved):

PR #282, opened 2021-05-06
apitest recognize Set-Cookie header and set cookie to latter apis

sfwn:
#### What type of PR is this?

bug

use `Set-Cookie` header in response instead of CookieJar, and set `Cookie` header for latter apis to simulate login.

### DIFF
--- a/modules/pipeline/aop/plugins/task/plugins/autotest_cookie_keep_before/plugin.go
+++ b/modules/pipeline/aop/plugins/task/plugins/autotest_cookie_keep_before/plugin.go
@@ -15,12 +15,15 @@ package autotest_cookie_keep_before
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/erda-project/erda/apistructs"
 	"github.com/erda-project/erda/modules/pipeline/aop/aoptypes"
 	"github.com/erda-project/erda/modules/pipeline/aop/plugins/task/plugins/autotest_cookie_keep_after"
+	"github.com/erda-project/erda/modules/pipeline/pipengine/reconciler/rlog"
+	"github.com/erda-project/erda/pkg/apitestsv2"
 )
 
 const taskType = "api-test"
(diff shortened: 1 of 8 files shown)


modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/meta.go
```diff
@@ -86,11 +88,8 @@ func writeMetaFile(ctx context.Context, task *spec.PipelineTask, meta *Meta) {
 	}
 	if meta.Resp != nil {
 		kvs.add(metaKeyAPIResponse, jsonOneLine(ctx, meta.Resp))
-	}
-	if meta.CookieJar != nil {
-		if meta.Resp != nil && meta.Resp.Headers != nil && meta.Resp.Headers["Set-Cookie"] != nil {
-			jar, _ := json.Marshal(meta.CookieJar)
-			kvs.add(metaKeyAPICookies, string(jar))
+		if headerSetCookie := meta.Resp.Headers[apitestsv2.HeaderSetCookie]; len(headerSetCookie) > 0 {
```
kakj-go, line 91: if Resp is nil,  Resp.Header will npe

modules/pipeline/aop/plugins/task/plugins/autotest_cookie_keep_before/plugin.go
```diff
@@ -44,45 +47,58 @@ func (p *Plugin) Handle(ctx *aoptypes.TuneContext) error {
...
+			rlog.TErrorf(ctx.SDK.Pipeline.ID, ctx.SDK.Task.ID, "failed to unmarshal AUTOTEST_API_GLOBAL_CONFIG, err: %v", err)
 			return err
 		}
 	}
 	if config.Header == nil {
 		config.Header = map[string]string{}
 	}
-	config.Header[autotest_cookie_keep_after.CookieJar] = cookieJar
-	configJson, err := json.Marshal(config)
+	config.Header[apitestsv2.HeaderCookie] = setCookie
```
kakj-go, line 88: if config user set cookie in header, here you will cover user cookie

modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/meta.go
sfwn: Yes, I will check it.

modules/pipeline/aop/plugins/task/plugins/autotest_cookie_keep_before/plugin.go
sfwn: Overwrite cookie can implement user switch.

johnlanni (approved):

PR #285, opened 2021-05-06
Metrics query unit test coverage optimize

luoyuctl:
What type of PR is this?

- Complete unit test of custom function without environment dependency in SELECT.

/kind unit test


luobily (approved):
lgtm

PR #292, opened 2021-05-07
refactor startup

recallsong:
#### What type of PR is this?

/kind refactor

refactor startup

#### Specified Reivewers:
/assgin @johnlanni @Effet @luobily


modules/volume-provisioner/provider.go
```diff
@@ -17,41 +17,10 @@ package volumeprovisioner
 import (
 	"context"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/erda-project/erda-infra/base/servicehub"
+	"github.com/sirupsen/logrus"
```
luobily, line 21:
```suggestion
	"context"

	"github.com/sirupsen/logrus"

	"github.com/erda-project/erda-infra/base/servicehub"
```

luobily: the order of imports:
- Standard library imports
- Third party imports,
- Own imports.

luobily (changes requested):

modules/volume-provisioner/provider.go
recallsong: the order of imports:
1、Standard package imports
2、Other imports
it is enough, and not necessary to distinguish between Third party imports and Own imports.

liuhaoyang (approved):

luobily (approved):
lgtm

PR #299, opened 2021-05-08
batch insert pipeline

kakj-go:
#### What type of this PR
feature

If the pipeline has a lot of  snippet, it turns out that one task and one pipeline are gradually created in this way will cause the request to time out, the modification here is gradually created as a batch creation


modules/qa/services/autotest_v2/testplan_v2.go
```diff
@@ -416,15 +416,10 @@ func (svc *Service) ExecuteDiceAutotestTestPlan(req apistructs.AutotestExecuteTe
...
-	resp, err := svc.bdl.GetTestPlanV2(req.TestPlan.ID)
-	if err != nil {
-		return nil, err
-	}
-
 	var spec pipelineyml.Spec
 	spec.Version = "1.1"
 	var stagesValue []*pipelineyml.Stage
-	for _, v := range resp.Data.Steps {
+	for _, v := range testPlan.Steps {
```
sfwn, line 422: move these to another pr.

sfwn (changes requested):

sfwn:
Did you test the condition: shutdown pipeline service while batch creating process? 
After reboot, the corrupted data is broken or compatible for present.

sfwn:
Bad implementation, it's difficult to maintain code.
Please consider implement it at ORM level.

sfwn:
I will close it.
You can reopen it once you adjusted.

PR #300, opened 2021-05-08
disable ess by default

WeiXuSeu:
#### What type of this PR

/kind optimize

disable ess by default

no consumer user is now, and it generate many unuseful log

#### Specified Reivewers:
/assgin @luobily


modules/ops/conf/conf.go
```diff
@@ -21,6 +21,7 @@ import (
 // Conf Define the configuration
 type Conf struct {
 	Debug         bool   `env:"DEBUG" default:"false"`
+	EnableEss     bool   `env:"ENABLE_ESS" default:"false"`
```
luobily, line 24: What scenarios need to be opened? Where to configure?

WeiXuSeu: It's a ecs auto scaling feature come from exxonmobil, but not used, and can be enabled by dice yaml environment.

luobily (commented):
lgtm

luobily (approved):

PR #308, opened 2021-05-10
Topology bug fix

luoyuctl:
#### What type of this PR
Topology bug fix

/kind bug

Fixes #
- Solve the problem of overcalculating the number of application instances
- Fixed an issue where the running count of application instances on the topology page was incorrectly displayed 
- Fixed the incorrect number of incorrect requests and service exceptions in the global overview

### DIFF
--- a/modules/monitor/apm/topology/topology.go
+++ b/modules/monitor/apm/topology/topology.go
@@ -1145,7 +1145,7 @@ func (topology *provider) GetOverview(language i18n.LanguageCodes, params Global
 
 	// error request count
 	errorCount := 0.0
-	for _, errorReqMetricName := range ErrorReqMetricNames {
+	for _, errorReqMetricName := range ReqMetricNames {
 		count, err := topology.globalReqCount(errorReqMetricName, params, metricsParams)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


modules/monitor/apm/topology/topology.go
```diff
@@ -1358,6 +1356,7 @@ func (topology *provider) Services(serviceName string, nodes []*Node) []ServiceD
 
 type ServiceInstance struct {
 	ApplicationName     string `json:"applicationName,omitempty"`
+	ServiceId           string `json:"serviceId,omitempty"`
```
luobily, line 1359: Proper nouns need to be capitalized, e.g. ID

luobily (changes requested):

modules/monitor/apm/topology/topology.go
luoyuctl: Combining id with other fields is universal, and id as a separate field needs to be all uppercase

luoyuctl: > Proper nouns need to be capitalized, e.g. ID
Combining id with other fields is universal, and id as a separate field needs to be all uppercase

luoyuctl: > Proper nouns need to be capitalized, e.g. ID

Combining id with other fields is universal, and id as a separate field needs to be all uppercase

liuhaoyang (approved):

PR #340, opened 2021-05-13
openapi oauth2 disallowed token request without accessTokenExpiredIn param

sfwn:
#### What type of this PR

feature

Openapi oauth2 disallowed token request without accessTokenExpiredIn param.
Client must set expected accessTokenExpiredIn explicitly.
All related clients have been adjusted.

### DIFF
--- a/modules/openapi/oauth2/jwt.go
+++ b/modules/openapi/oauth2/jwt.go
@@ -75,7 +75,9 @@ func (a *JWTAccessGenerate) Token(data *oauth2.GenerateBasic, isGenRefresh bool)
 	}
 
 	// expire_in
-	if payload.AccessTokenExpiredIn != "" {
+	if payload.AccessTokenExpiredIn == "" {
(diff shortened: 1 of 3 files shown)


modules/openapi/oauth2/jwt.go
```diff
@@ -75,7 +75,9 @@ func (a *JWTAccessGenerate) Token(data *oauth2.GenerateBasic, isGenRefresh bool)
 	}
 
 	// expire_in
-	if payload.AccessTokenExpiredIn != "" {
+	if payload.AccessTokenExpiredIn == "" {
+		return "", "", fmt.Errorf("no accessTokenExpiredIn is not allowed")
```
Effet, line 79:
```suggestion
		return "", "", fmt.Errorf("accessTokenExpiredIn is empty, which is not allowed.")
```

Effet (changes requested):

Effet (approved):

PR #355, opened 2021-05-14
generate and replace random value before snippetTask created

chengjoey:
#### What type of this PR

/kind bug

generate and replace random value before snippet task created, and so actions can use the same random param value

#### Specified Reivewers:
/assgin @your-reviewer


sfwn (changes requested):
How to achieve global mock unification at the pipeline level, not only at snippetLevel

PR #358, opened 2021-05-17
CalculatePipelineTaskAllDone mean all task was end, bug now if have one task status is error function will return true

kakj-go:
#### What type of this PR
 feature

CalculatePipelineTaskAllDone mean all task was end, bug now if have one task status is error function will return true

### DIFF
--- a/modules/pipeline/commonutil/statusutil/status.go
+++ b/modules/pipeline/commonutil/statusutil/status.go
@@ -82,7 +82,16 @@ func CalculatePipelineStatusV2(tasks []*spec.PipelineTask) apistructs.PipelineSt
 	}
 }
 
-// CalculatePipelineTaskAllDone 计算 pipeline 下的所有任务是否都已完毕
+// CalculatePipelineTaskAllDone
+// all task was Disabled or EndStatus, return true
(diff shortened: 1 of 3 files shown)


modules/pipeline/commonutil/statusutil/status.go
```diff
@@ -82,7 +87,16 @@ func CalculatePipelineStatusV2(tasks []*spec.PipelineTask) apistructs.PipelineSt
 	}
 }
 
-// CalculatePipelineTaskAllDone 计算 pipeline 下的所有任务是否都已完毕
+// CalculatePipelineTaskAllDone
+// all task is Success Disabled or Failed, return true
```
sfwn, line 91: task is Disabled or EndStatus

modules/pipeline/commonutil/statusutil/status.go
```diff
@@ -19,6 +19,11 @@ import (
 )
 
 // CalculatePipelineStatusV2
+// if have task was Failed that pipeline was Failed
```
sfwn, line 22: Adjust these comments please.

sfwn (changes requested):

sfwn (approved):

PR #359, opened 2021-05-17
modify streaming.yml add kafka topics

Counterflowwind:
/fix

mysql connection time out

#### Specified Reivewers:
/assgin @liuhaoyang


bundle/urls.go
```diff
@@ -90,7 +90,8 @@ func (u urls) Collector() (string, error) {
 }
 
 func (u urls) Monitor() (string, error) {
-	return u.getURL(discover.EnvMonitor, discover.SvcMonitor)
+	//return u.getURL(discover.EnvMonitor, discover.SvcMonitor)
+	return "localhost:7096", nil
```
luoyuctl, line 94: why use the hard coded? 'localhost:7096' wasn't the generic address.

luoyuctl (approved):

bundle/urls.go
luoyuctl, line 94: why use the hard coded? 'localhost:7096' wasn't the generic address.

luoyuctl (changes requested):

liuhaoyang (approved):

PR #363, opened 2021-05-17
SearchExtensions filter same action type

kakj-go:
#### What type of this PR

/kind feature

SearchExtensions filter same action type


sfwn (changes requested):
use strutil.dedup

sfwn (approved):

Effet (approved):

PR #372, opened 2021-05-18
Fix/compoment add runuser

chengjoey:
#### What type of this PR

/kind bug

autotest detail add pipeline run user info


sfwn:
make commits verified please.

Effet (approved):

PR #378, opened 2021-05-19
ignore not found instance when destory runtime

Muzry:
#### What type of this PR

/kind bug

- ignore not found instance when destory runtime

Fixes #

#### Specified Reivewers:
/assgin @luobily


modules/scheduler/executor/plugins/k8s/runtime.go
```diff
@@ -98,7 +98,7 @@ func (k *Kubernetes) destroyRuntimeByProjectNamespace(ns string, sg *apistructs.
 		default:
 			err = k.deleteDeployment(ns, service.Name)
 		}
-		if err != nil {
+		if err != nil && !strings.Contains(err.Error(), "not found") {
```
luobily, line 101:
```suggestion
		if err != nil && !apierrors.IsNotFound(err) {
```

luobily: This judgment is useful in many places, add a function to use.

luobily (changes requested):

PR #380, opened 2021-05-19
optimize file upload to avoid OOM

littlejiancc:
#### What type of this PR
/kind bug
/kind bug
/kind cleanup
/kind documentation
/kind feature
/kind design

optimize file upload to avoid OOM

Fixes #

#### Specified Reivewers:
/assgin @your-reviewer


modules/cmdb/endpoints/file.go
```diff
@@ -42,29 +42,28 @@ const (
...
-
-	// 获取上传文件
 	formFile, fileHeader, err := r.FormFile("file")
 	if err != nil {
 		return nil, apierrors.ErrUploadFile.InternalError(err)
 	}
 	defer formFile.Close()
 
-	// 获取参数
+	// check the size
```
sfwn, line 61: check file size should at very beginning.

sfwn (changes requested):

modules/cmdb/endpoints/file.go
littlejiancc: now use the content length to check file size

sfwn (dismissed):

sfwn (approved):

PR #381, opened 2021-05-19
modify loop calculate interval to make interval is accurate

chengjoey:
#### What type of this PR

/kind bug
loop calculate interval change looped times + 1 to looped times - 1


sfwn (approved):

sfwn:
will be merge at v1.1

PR #394, opened 2021-05-19
pipeline apitest use json oneline instead of strutil string

chengjoey:
#### What type of this PR

/kind bug

use jsonOnline instead of strutil string compatible json type param, header, assert

### DIFF
--- a/modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/envconf.go
+++ b/modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/envconf.go
@@ -15,7 +15,7 @@ package logic
 
 import (
 	"github.com/erda-project/erda/apistructs"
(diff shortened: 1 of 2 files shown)


modules/pipeline/pipengine/actionexecutor/plugins/apitest/logic/envconf.go
```diff
@@ -15,7 +15,7 @@ package logic
 
 import (
 	"github.com/erda-project/erda/apistructs"
-	"github.com/erda-project/erda/pkg/strutil"
+	"github.com/erda-project/erda/pkg/jsonparse"
```
sfwn, line 18: add unit test for jsonparse.JsonOneline func.

sfwn (changes requested):

sfwn (approved):

PR #434, opened 2021-05-25
autotest move scene check in same sceneSet not have same pre_id scene

kakj-go:
#### What type of this PR

/kind feature

autotest move scene check in same sceneSet not have same pre_id scene


modules/qa/dao/autotest_scene.go
```diff
@@ -160,9 +160,16 @@ func (db *DBClient) DeleteAutoTestScene(id uint64) error {
 		if err := tx.Save(&next).Error; err != nil {
 			return err
 		}
+
+		defer func() {
```
sfwn, line 164: err won't be returned

sfwn (changes requested):

sfwn (approved):

PR #467, opened 2021-05-27
fix(scheduler): leak namespace

Muzry:
#### What type of this PR
/kind bug

- fix(scheduler): leak namespace

Fixes #

#### Specified Reviewers:
/assign @luobily


luobily (approved):
lgtm

PR #485, opened 2021-05-28
fix: change map to sync map to avoid concurrent problem

littlejiancc:
#### What type of this PR
/kind bug
fix: change map to sync map to avoid concurrent problem

https://erda-org.erda.cloud/erda/workBench/projects/387/issues/all?id=63187&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMTI2MSJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjF9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6InRhYmxlIiwiY2hpbGRyZW5WYWx1ZSI6eyJrYW5iYW4iOiJkZWFkbGluZSJ9fQ%3D%3D&iterationID=-1&type=BUG

#### Specified Reviewers:
/assign @QvodSoldier


modules/orchestrator/endpoints/addon_test.go
```diff
@@ -0,0 +1,201 @@
...
+			var ext []apistructs.Extension
+			for i, v := range keys {
+				ext = append(ext, apistructs.Extension{ID: uint64(i), Name: v})
+			}
+			return ext, nil
+		},
+	)
+	var (
+		wg sync.WaitGroup
+		n  = 2000
```
QvodSoldier, line 45: please reduce it，unit test no need for so much gorouting

modules/orchestrator/endpoints/addon_test.go
```diff
@@ -0,0 +1,201 @@
...
+		}()
+	}
+	wg.Wait()
+	for _, v := range keys {
+		if _, ok := addon.AddonInfos.Load(v); !ok {
+			t.Fatal("missing key")
+		}
+	}
+}
+
```
QvodSoldier, line 66: I think we should simulate the situation of one write and multiple reads.

modules/orchestrator/endpoints/addon_test.go
littlejiancc: now change to 20

QvodSoldier (dismissed):

PR #509, opened 2021-05-31
test plan generate performance optimization

chengjoey:
#### What type of this PR

/kind bug

Concurrent list apis for generate report performance optimization


modules/qa/services/testplan/report.go
```diff
@@ -38,24 +41,35 @@ func (t *TestPlan) GenerateReport(testPlanID uint64) (*apistructs.TestPlanReport
 		return nil, apierrors.ErrPagingTestPlanCaseRels.InternalError(err)
 	}
 	var totalApiCount apistructs.TestCaseAPICount
+	var relErr error
+	var wg sync.WaitGroup
 	for _, rel := range rels {
```
sfwn, line 46: You cannot async handle all rels at once. You should have rate limited.

sfwn (changes requested):

modules/qa/services/testplan/report.go
```diff
@@ -38,24 +41,43 @@ func (t *TestPlan) GenerateReport(testPlanID uint64) (*apistructs.TestPlanReport
...
+				case apistructs.ApiTestRunning:
+					totalApiCount.Running++
+				case apistructs.ApiTestPassed:
+					totalApiCount.Passed++
+				case apistructs.ApiTestFailed:
+					totalApiCount.Failed++
+				}
+			}
+			<-caseChan
+			defer wg.Done()
```
sfwn, line 73: defer should be at very beginning of func.

sfwn (approved):

PR #522, opened 2021-06-01
feat: first user become admin, only if kratos enabled

Effet:
#### What type of this PR

/kind feature

work with kratos, make first user as admin

Fixes #

#### Specified Reviewers:
/assign @sfwn

### DIFF
--- a/modules/cmdb/services/member/member.go
+++ b/modules/cmdb/services/member/member.go
@@ -25,6 +25,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/erda-project/erda/apistructs"
+	"github.com/erda-project/erda/modules/cmdb/conf"
 	"github.com/erda-project/erda/modules/cmdb/dao"
 	"github.com/erda-project/erda/modules/cmdb/model"
 	"github.com/erda-project/erda/modules/cmdb/services/apierrors"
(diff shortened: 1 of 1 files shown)


modules/cmdb/services/member/member.go
```diff
@@ -728,9 +728,9 @@ func (m *Member) IsAdmin(userID string) bool {
 		}
 	}
 	if member.ID == 0 {
-		logrus.Warnf("CAUTION: user(%s) current no admin, will become soon", userID)
+		logrus.Warnf("CAUTION: user(%s) currently not an admin, will become soon if no one admin exist", userID)
 		// TODO: some risk
-		if m.noOneAdmin() && len(userID) > 11 {
+		if len(userID) > 11 && m.noOneAdminForKratos() { // len > 11 imply that is kratos user
```
sfwn, line 733: why 11

Effet: just a magic value, kratos' user_id is UUID, it is significantly larger than 11

sfwn: please comment it clearly for latter coders.

sfwn: please comment it clearly for latter coders.

sfwn (changes requested):

modules/cmdb/services/member/member.go
Effet: you're right, added

sfwn (approved):

PR #540, opened 2021-06-03
feature(pipeline): queue polish the logic to set needReRangePendingQueue flag

sfwn:
#### What type of this PR

feature

Queue set needReRangePendingQueue flag if newItem's order is higher than the current ranging one.

### DIFF
--- a/modules/pipeline/pipengine/queue/priorityqueue/priority_queue.go
+++ b/modules/pipeline/pipengine/queue/priorityqueue/priority_queue.go
@@ -94,3 +94,21 @@ func (pq *PriorityQueue) Range(f func(Item) (stopRange bool)) {
 		}
 	}
 }
+
+// LeftHasHigherOrder judge order of two items.
+// return true if left has higher order.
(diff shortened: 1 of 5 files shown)


modules/pipeline/pipengine/queue/priorityqueue/priority_queue.go
```diff
@@ -94,3 +94,21 @@ func (pq *PriorityQueue) Range(f func(Item) (stopRange bool)) {
...
+		return true
+	}
+	if !leftExist {
+		return false
+	}
+	if !rightExist {
+		return true
+	}
+	// both exist, judge by index
+	return leftItem.Index() < rightItem.Index()
```
kakj-go, line 113: only leftItem.Index() < rightItem.Index() return true other return false

kakj-go (approved):

modules/pipeline/pipengine/queue/priorityqueue/priority_queue.go
sfwn: if right item doesn't exists, return true.

PR #550, opened 2021-06-03
Feature/cron metric issue

kakj-go:
#### What type of this PR
/kind feature

Feature/cron metric issue


apistructs/issue.go
```diff
@@ -686,6 +686,8 @@ type IssueListRequest struct {
 	IdentityInfo
 	// 用来区分是通过ui还是bundle创建的
 	External bool `json:"-"`
+	// simple
+	SimpleResult bool `json:"simpleResult"`
```
sfwn, line 690: No one knows how many fields will be returned in Simple-Mode.
Maybe `OnlyID` is better.

sfwn (changes requested):

sfwn (approved):

PR #553, opened 2021-06-03
Reconstruction anomaly analysis

luoyuctl:
#### What type of this PR

/kind bug

Switch the exception analysis list from looking up Cassandra to ElasticSearch, and adjust part of the logic.

Fixed the problem of occasionally querying invalid empty data to optimize performance.


innerpeacez (approved):
/lgtm

PR #587, opened 2021-06-09
permission interceptor

recallsong:
#### What type of this PR
/kind polish

1、permission interceptor

#### Specified Reviewers:
@liuhaoyang 

no


pkg/common/permission/permission_test.go
```diff
@@ -0,0 +1,116 @@
...
+			args: args{
+				method: testInterface.TestFunc_1,
+			},
+			want: "TestFunc_1",
+		},
+		{
+			args: args{
+				method: testInterface.函数_1,
+			},
+			want: "函数_1",
```
luoyuctl, line 98: why use cn case?

PR #599, opened 2021-06-10
render return error then continue render info and show error info

kakj-go:
#### What type of this PR
bugfix

/kind bug

render return error then continue render info and show error info


modules/openapi/api/apis/openapi/component_protocol/render.go
```diff
@@ -122,7 +122,7 @@ func RenderResponse(req *apistructs.ComponentProtocolRequest) *apistructs.Compon
 	}
 
 	err := protocol.GetGlobalStateKV(req.Protocol, protocol.GlobalInnerKeyError.String())
-	if err != nil {
+	if err != nil && len(err.(string)) > 0 {
```
sfwn, line 125: err.(string) is unsafe.

sfwn (changes requested):

sfwn (approved):

PR #618, opened 2021-06-16
Fix the problem that the forced conversion may be abnormal

luoyuctl:
#### What type of this PR

/kind bug

- Fixes #66559
- [Erda Cloud Issue Link](Fix the problem that the forced conversion may be abnormal )


recallsong (approved):
/approve

PR #635, opened 2021-06-21
pipeline run user use nick info not user name

kakj-go:
#### What type of this PR
/kind feature

pipeline run user use nick info not user name,  user name was not meaningful fields


apistructs/pipeline_dto.go
```diff
@@ -184,7 +184,7 @@ type PipelineStageDetailDTO struct {
 func (user *UserInfo) ConvertToPipelineUser() *PipelineUser {
 	return &PipelineUser{
 		ID:     user.ID,
-		Name:   user.Name,
+		Name:   user.Nick,
```
sfwn, line 187: You should assign a value if Nick is empty.

sfwn (changes requested):

sfwn (approved):

PR #638, opened 2021-06-21
permission code refactor, app scopeType permission extension project-app permission

kakj-go:
#### What type of this PR

/kind feature

in project level app, if project level app have permission,  other app should have same permission. only in this way can have a good interactive experience


modules/cmdb/endpoints/permission.go
```diff
@@ -231,102 +230,20 @@ func (e *Endpoints) getPermission(userID string, scopeType apistructs.ScopeType,
...
-		}
-		resourceRoleList = append(resourceRoleList, rr)
-	}
+	permissionAdaptor.Once.Do(func() {
+		permissionAdaptor.Db = e.db
+		permissionAdaptor.Bdl = e.bdl
+		permissionAdaptor.RegisterHandler(security.AdminUserPermissionHandler{Adaptor: &permissionAdaptor})
+		permissionAdaptor.RegisterHandler(security.SupportUserPermissionHandler{Adaptor: &permissionAdaptor})
+		permissionAdaptor.RegisterHandler(security.NormalUserPermissionHandler{Adaptor: &permissionAdaptor})
+	})
```
Effet, line 244: is it better initialize security.PermissionAdaptor as service in initialize.go? cc @sfwn

modules/cmdb/security/adaptor.go
```diff
@@ -0,0 +1,178 @@
...
+				return apistructs.PermissionList{
+					Access:           false,
+					PermissionList:   make([]apistructs.ScopeResource, 0),
+					ResourceRoleList: make([]apistructs.ScopeResource, 0),
+					Roles:            make([]string, 0),
+					Exist:            true,
+				}, err
+			}
+
+			return *list, err
```
Effet, line 167: get all permission differ than check permission.

it should not just return only one handler accessed, it maybe accessed as low-permission role, we shall combine all accessed handler's role/permission.

Effet (changes requested):

modules/dop/endpoints/pipeline.go
```diff
@@ -58,6 +59,10 @@ func (e *Endpoints) pipelineCreate(ctx context.Context, r *http.Request, vars ma
 		if err := e.permission.CheckRuntimeBranch(identityInfo, createReq.AppID, createReq.Branch, apistructs.OperateAction); err != nil {
 			return errorresp.ErrResp(err)
 		}
+	} else if len(r.Header.Get(httputil.InternalActionHeader)) > 0 {
+		if err := e.permission.CheckRuntimeBranch(identityInfo, createReq.AppID, createReq.Branch, apistructs.OperateAction); err != nil {
+			return errorresp.ErrResp(err)
```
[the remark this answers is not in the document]
kakj-go: some action Internal request should check permission,  like create app or run pipeline

modules/cmdb/endpoints/permission.go
kakj-go: it is found that the two processing logics of check and access are somewhat different, so the permission processor in the adaptor is dynamically configurable for each interface

You can find two places use adapter, and their number of processors is different

modules/cmdb/security/adaptor.go
kakj-go: if list is empty or has error, i think it should return empty permission, like old logic

modules/cmdb/endpoints/permission.go
Effet: i know then, but it can still be initialized as two service?

pkg/http/httputil/header.go
```diff
@@ -15,10 +15,11 @@ package httputil
...
 // dice 公共 Header
 const (
-	UserHeader      = "User-ID"
-	OrgHeader       = "Org-ID"
-	InternalHeader  = "Internal-Client" // 内部服务间调用时使用
-	RequestIDHeader = "RequestID"
+	UserHeader           = "User-ID"
+	OrgHeader            = "Org-ID"
+	InternalHeader       = "Internal-Client"        // 内部服务间调用时使用
+	InternalActionHeader = "Internal-Action-Client" // 内部间调用时使用
```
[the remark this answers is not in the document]
kakj-go: some action Internal request should check permission, like create app or run pipeline

Effet (approved):

PR #639, opened 2021-06-21
move tmc functions to erda

Counterflowwind:
What type of this PR
feature

move tmc functions to erda

Specified Reviewers:
/assign @liuhaoyang


luoyuctl (commented):
for title , use migrated xx module

PR #641, opened 2021-06-21
Feature/dop refactor

littlejiancc:
#### What type of this PR
/kind feature

dop and core-services refactor


sfwn:
After refactor, unit test Coverage rate from 10.65%  to 10.09%, -0.57%.

apistructs/application.go
```diff
@@ -192,6 +192,8 @@ type ApplicationDTO struct {
 
 	// 应用更新时间
 	UpdatedAt time.Time `json:"updatedAt"`
+
+	Extra string
```
sfwn, line 196: Remove this useless field.

sfwn (changes requested):

apistructs/application.go
littlejiancc: update to 
``` 
Extra string `json:"-"`
```

sfwn (approved):

PR #653, opened 2021-06-23
Refactored metrics module meta interface to protobuf

luoyuctl:
#### What type of this PR

/kind feature

Using protobuf to generate HTTP interface to replace the original HTTP interface.

All interfaces have been verified.

image


luoyuctl:
GPG key was deleted , commit unverified ,close it.

PR #682, opened 2021-06-25
Test case file records table migration

shuofan:
#### What type of this PR
feature

Test case file records table migration

#### Specified Reviewers:

/assign @sfwn @dspo


sfwn:
This sql migration file should under the `qa` or new `dop` ?

sfwn (approved):

PR #688, opened 2021-06-25
use pkg/registryhelper instead of remote soldier access

johnlanni:
#### What type of this PR

/kind feature

use pkg/registryhelper instead of remote soldier access


modules/dicehub/registry/manifests.go
```diff
@@ -15,41 +15,36 @@
...
-		Do().
-		JSON(&imageResp)
+	clusterInfo, err := bdl.QueryClusterInfo(clusterName)
 	if err != nil {
-		return errors.Errorf("recycle image: %+v error: %v", images, err)
+		return err
 	}
-	if !resp.IsOK() || !imageResp.Success {
-		return errors.Errorf("recycle image: %+v fail, statusCode: %d, err: %+v", images, resp.StatusCode(), imageResp.Error)
+	registryUrl := clusterInfo.MustGet(apistructs.REGISTRY_ADDR)
```
sfwn, line 37: MustGet will panic if registryUrl is empty.

sfwn (changes requested):

sfwn (approved):

PR #693, opened 2021-06-28
put the loop strategy in the framework.go, the task during the loop no longer has a end state at the databases, and it is directly updated to the analyzed state after the loop sleep

kakj-go:
/kind bug

If two tasks are executed at the same time, and one task has a loop configuration. The pipeline should wait for the end of the loop task, but because the waiting task will also be in the end state, this will cause the pipeline to misjudge that all the tasks are over, thereby updating its state to the end state, which will cause the task to terminate the loop


modules/pipeline/pipengine/reconciler/taskrun/framework.go
```diff
@@ -180,3 +188,85 @@ func (tr *TaskRun) LogStep(taskOp Op, step string) {
 	logrus.Debugf("reconciler: pipelineID: %d, taskID: %d, taskName: %s, taskOp: %s, step: %s",
 		tr.P.ID, tr.Task.ID, tr.Task.Name, string(taskOp), step)
 }
+
```
sfwn, line 191: Just move old file to new path: reconciler/taskrun/task_loop.go

sfwn (changes requested):

sfwn (approved):

PR #709, opened 2021-06-29
Refactor metric native query to protobuf type

luoyuctl:
#### What type of this PR

/kind feature

Refactored metrics native query apis , the interface output is compatible with each other.

The interface has been tested and the entry and exit parameters are normal.
image

#### Specified Reviewers:

/assign @recallsong


luoyuctl:
Conflict with #689 , close.

PR #751, opened 2021-07-02
Add test api file size, stage value not null

shuofan:
#### What type of this PR
feature

Add test api file size, custom issue stage value not null

#### Specified Reviewers:

/assign @Effet @sfwn


modules/dop/services/testcase/export.go
```diff
@@ -174,8 +174,15 @@ func (svc *Service) ExportFile(record *dao.TestFileRecord) {
 			return
 		}
 
+		fileInfo, err := f.Stat()
+		if err != nil {
+			logrus.Error(apierrors.ErrExportTestCases.InternalError(err))
```
Effet, line 179: log should be more readable

Effet (changes requested):

modules/dop/services/testcase/export.go
shuofan: Done

Effet (approved):

PR #768, opened 2021-07-05
task prepare use k8sjob executor create volume

chengjoey:
#### What type of this PR

/kind feature

use k8sjob executor create job volume instead of call scheduler

### DIFF
--- a/modules/pipeline/pipengine/reconciler/taskrun/taskop/prepare.go
+++ b/modules/pipeline/pipengine/reconciler/taskrun/taskop/prepare.go
@@ -31,6 +31,8 @@ import (
 	"github.com/erda-project/erda/modules/actionagent"
 	"github.com/erda-project/erda/modules/pipeline/aop/aoptypes"
 	"github.com/erda-project/erda/modules/pipeline/conf"
+	"github.com/erda-project/erda/modules/pipeline/pipengine/actionexecutor/plugins/scheduler"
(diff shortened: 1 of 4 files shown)


modules/pipeline/pipengine/reconciler/taskrun/taskop/prepare.go
```diff
@@ -430,10 +432,26 @@ func (pre *prepare) makeTaskRun() (needRetry bool, err error) {
...
+		schedExecutor, ok := pre.Executor.(*scheduler.Sched)
+		if !ok {
+			return false, fmt.Errorf("failed to createJobVolume, err: invalid task executor kind")
+		}
+		schedPlugin, err := schedExecutor.GetTaskExecutor(task.Type, p.ClusterName, task)
+		if err != nil {
+			return false, fmt.Errorf("failed to createJobVolume, err: can not get k8s executor")
+		}
+		if schedPlugin.Kind() != k8sjob.Kind {
+			return false, fmt.Errorf("can not createJobVolume, err: invalid task executor")
```
sfwn, line 445: continue

sfwn (changes requested):

modules/pipeline/pipengine/reconciler/taskrun/taskop/prepare.go
sfwn: task executor %s not support share volume

sfwn (changes requested):

sfwn (approved):

PR #782, opened 2021-07-07
update dop client and pipeline template sql

kakj-go:
#### What type of this PR
/kind feature

1. because qa and gittar-adaptor are merged into the dop module, some snippet config clients need to change the address to dop, direct update does not determine whether it exists, because pipeline-base.sql is data created with a fixed id
2. pipeline template sql have some problems，wrong line break


dspo:
please do not alter baseline

PR #839, opened 2021-07-09
Fix the problem that dialer does not transparently transmit backend response headers

johnlanni:
…esponse headers

#### What type of this PR

/kind bug

Fix the problem that dialer does not transparently transmit backend response headers


snakorse (approved):

iutx (approved):
lgtm @luobily

PR #842, opened 2021-07-09
modify create org regex, add default if not planFinishedAt

chengjoey:
#### What type of this PR

/kind bug

modify org name regex, add default value if planFinishedAt is nil

### DIFF
--- a/modules/core-services/utils/regexp.go
+++ b/modules/core-services/utils/regexp.go
@@ -15,7 +15,8 @@ package utils
 
 import "regexp"
 
-var orgReg = regexp.MustCompile(`^[0-9a-z]\d*[a-z-][a-z\d-]*[0-9a-z]$`)
+// org regexp modify from pkg/strutil/regexp.go/reg, but org name can not be pure digital
+var orgReg = regexp.MustCompile(`^(?:[a-z]+|[0-9]+[a-z]+|[0-9]+[-]+[a-z0-9])+(?:(?:(?:[-]*)[a-z0-9]+)+)?$`)
 
 // IsValidOrgName check org name can contain a-z0-9- but can not pure 0-9
 func IsValidOrgName(repo string) bool {
--- a/modules/openapi/component-protocol/scenarios/home-page-content/components/tableGroup/render.go
+++ b/modules/openapi/component-protocol/scenarios/home-page-content/components/tableGroup/render.go
@@ -270,6 +270,7 @@ func (t *TableGroup) addWorkbenchData(datas *apistructs.WorkbenchResponse, orgNa
 				},
 				OrgName: orgName,
 			}
+			issueItem.PlanFinishedAt = i18nLocale.Get(i18n.I18nKeyIssueNotSpecifiedDate)
 			if issue.PlanFinishedAt != nil {
 				issueItem.PlanFinishedAt = issue.PlanFinishedAt.Format("2006-01-02")
 			}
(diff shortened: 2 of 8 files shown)


modules/openapi/component-protocol/scenarios/home-page-content/components/tableGroup/render.go
```diff
@@ -270,6 +270,7 @@ func (t *TableGroup) addWorkbenchData(datas *apistructs.WorkbenchResponse, orgNa
 				},
 				OrgName: orgName,
 			}
+			issueItem.PlanFinishedAt = i18nLocale.Get(i18n.I18nKeyIssueNotSpecifiedDate)
```
chengjoey, line 273: image

pkg/erda-configs/i18n/workbench.json
```diff
@@ -79,7 +79,8 @@
     "wb.content.component.issue.tomorrowExpired": "，明天到期:",
     "wb.content.component.issue.sevenDayExpired": "，7日内到期:",
     "wb.content.component.issue.thirtyDayExpired": "，30日内到期:",
-    "wb.content.component.issue.featureExpired": "，未来:"
+    "wb.content.component.issue.featureExpired": "，未来:",
+    "wb.content.component.issue.notSpecifiedDate": "未指定截止日期"
```
Effet, line 83: concat translated text is not a good practices, shall we support argument for i18n item, like:

`wb.content.component.issue.yourRemains: "当前你还有 %d，%d 个xxx，%d 个yyy"`

modules/core-services/utils/regexp.go
```diff
@@ -15,7 +15,7 @@ package utils
 
 import "regexp"
 
-var orgReg = regexp.MustCompile(`^[0-9a-z]\d*[a-z-][a-z\d-]*[0-9a-z]$`)
+var orgReg = regexp.MustCompile(`^(?:[a-z]+|[0-9]+[a-z]+|[0-9]+[-]+[a-z0-9])+(?:(?:(?:[-]*)[a-z0-9]+)+)?$`)
```
Effet, line 18: make a comment describe what this regexp means, it's too complex

modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
```diff
@@ -102,7 +102,7 @@ func (o *OrgFormModal) SetProps() {
 				"rules": []interface{}{
 					map[string]interface{}{
 						"msg":     "由小写字母、数字，-组成(不允许纯数字)",
-						"pattern": "/^(?![0-9]+$)[a-z0-9-]*$/",
+						"pattern": "/^(?![0-9]+$)^[^-](.*[^-])?$[a-z0-9-]*$/",
```
Effet, line 105: make a comment describe what this regexp means, it's too complex

Effet (changes requested):

pkg/erda-configs/i18n/workbench.json
chengjoey: these issue status used in different struct, so can`t make in one argument

modules/core-services/utils/regexp.go
chengjoey: org regex modify from 
image
,but org name can not be pure digital

Effet: add into code, and use english

modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
chengjoey: same with orgReg, but Different front-end writing

Effet (approved):

PR #873, opened 2021-07-12
autotest scene pre parallel steps output shouldn't show between parallel steps

kakj-go:
#### What type of this PR
/kind bug

autotest scene pre parallel steps output shouldn't show between parallel steps


modules/openapi/component-protocol/scenarios/auto-test-scenes/components/apiEditor/render.go
```diff
@@ -157,12 +157,12 @@ LABEL:
 		if sStep.ID == stepID {
 			break
 		}
-		steps = append(steps, sStep)
 		for _, pStep := range sStep.Children {
 			if pStep.ID == stepID {
 				break LABEL
 			}
 		}
+		steps = append(steps, sStep)
```
Effet, line 165: can you explain what logic in it?

Effet (changes requested):

modules/openapi/component-protocol/scenarios/auto-test-scenes/components/apiEditor/render.go
kakj-go: The old logic is to judge whether it is a parallel node first, and it should be a parallel node before attaching

sfwn (approved):

PR #910, opened 2021-07-13
collect metrics bdl use new httpclient

kakj-go:
#### What type of this PR
/kind bug

dop report issue data to collector, request will set Authorization header because basicAuth not check user name and password, it will make can not report data, collector response 401 status, The cause of the problem is that the same httpclient is used


pkg/http/httpclient/client.go
```diff
@@ -380,7 +380,7 @@ func (c *HTTPClient) Method(method string, host string, retry ...RetryOption) *R
 
 func (c *HTTPClient) newRequest(retry []RetryOption) *Request {
 	header := make(map[string]string)
-	if c.basicAuth != nil {
+	if c.basicAuth != nil && len(c.basicAuth.name) > 0 && len(c.basicAuth.password) > 0 {
```
sfwn, line 383: I think it's a bug of client side.
Password can be empty in BasicAuth.

sfwn (changes requested):

sfwn:
Change pr name and add test result please.

sfwn (approved):

PR #919, opened 2021-07-13
fix PipelineLink url

pipipipipi43:
#### What type of this PR

/kind bug

When there is a task that is already being executed in the pipeline, the link error of clicking the execution prompt is wrong
image
Steps to reproduce:
Corresponding to the existing branch of yml, click to create a new pipeline when executing, an error message is displayed, and the copy link cannot be opened

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/elastic/cloud-on-k8s v0.0.0-20210205172912-5ce0eca90c60
 	github.com/elazarl/goproxy v0.0.0-20200421181703-e76ad31c14f6
 	github.com/erda-project/erda-infra v0.0.0-20210709120204-7091e4656382
-	github.com/erda-project/erda-proto-go v0.0.0-20210714025152-31855363f907
+	github.com/erda-project/erda-proto-go v0.0.0-20210714112147-b4e7d90a9c31
 	github.com/extrame/ole2 v0.0.0-20160812065207-d69429661ad7 // indirect
 	github.com/extrame/xls v0.0.1
 	github.com/facebookgo/stack v0.0.0-20160209184415-751773369052 // indirect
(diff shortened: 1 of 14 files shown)


modules/pipeline/commonutil/linkutil/link.go
```diff
@@ -15,6 +15,7 @@ package linkutil
 
 import (
 	"fmt"
+	"github.com/erda-project/erda/modules/pipeline/conf"
```
Effet, line 18: import order

Effet (changes requested):

modules/pipeline/commonutil/linkutil/link.go
```diff
@@ -45,15 +46,7 @@ func GetPipelineLink(bdl *bundle.Bundle, p spec.Pipeline) (bool, string) {
...
-	// get domain protocol
-	clusterInfo, err := bdl.QueryClusterInfo(p.ClusterName)
-	if err != nil {
-		return false, ""
-	}
-	protocol := clusterInfo.Get(apistructs.DICE_PROTOCOL)
-	if protocol == "" {
-		return false, ""
-	}
+	Domain := conf.UIDomain()
```
kakj-go, line 49: Domain -> domain

modules/pipeline/commonutil/linkutil/link.go
```diff
@@ -45,15 +46,7 @@ func GetPipelineLink(bdl *bundle.Bundle, p spec.Pipeline) (bool, string) {
...
 
-	return true, fmt.Sprintf("%s://%s/workBench/projects/%d/apps/%d/pipeline/%d", protocol, org.Domain, projectID, appID, p.ID)
+	return true, fmt.Sprintf("https://%s/%s/dop/projects/%d/apps/%d/pipeline?pipelineID=%d", Domain, org.Name, projectID, appID, p.ID)
```
Effet, line 51: should move link fmt to dop.

Effet (changes requested):

modules/pipeline/commonutil/linkutil/link.go
Effet: and hard-code `https` is very bad

PR #923, opened 2021-07-13
modify orgFormModal org name regexp

chengjoey:
#### What type of this PR

/kind bug

modify the orgFormModal org`s name regexp
the previous regexp don`t match uppercase letter

### DIFF
--- a/modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
+++ b/modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
@@ -102,7 +102,7 @@ func (o *OrgFormModal) SetProps() {
 				"rules": []interface{}{
 					map[string]interface{}{
 						"msg":     "由小写字母、数字，-组成(不允许纯数字)",
-						"pattern": "/^(?![0-9]+$)^[^-](.*[^-])?$[a-z0-9-]*$/",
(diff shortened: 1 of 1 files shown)


modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
```diff
@@ -102,7 +102,7 @@ func (o *OrgFormModal) SetProps() {
 				"rules": []interface{}{
 					map[string]interface{}{
 						"msg":     "由小写字母、数字，-组成(不允许纯数字)",
-						"pattern": "/^(?![0-9]+$)^[^-](.*[^-])?$[a-z0-9-]*$/",
+						"pattern": "/^(?:[a-z]+|[0-9]+[a-z]+|[0-9]+[-]+[a-z0-9])+(?:(?:(?:[-]*)[a-z0-9]+)+)?$/",
```
sfwn, line 105: use const and add unit test for this.

sfwn (dismissed):

Effet (approved):

modules/openapi/component-protocol/scenarios/home-page-sidebar/components/orgFormModal/render.go
Effet: this regexp is for js, it's hard to unit test.

and we are planing disable frontend check, all check will in backend.

sfwn (approved):

PR #1025, opened 2021-07-15
update default bundle timeout

johnlanni:
#### What type of this PR

/kind feature

when we use telepresence in local, first dial is very slow, more than 5 seconds.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -46,7 +46,7 @@ func New(options ...Option) *Bundle {
 	}
 	if b.hc == nil {
 		b.hc = httpclient.New(
-			httpclient.WithTimeout(time.Second, time.Second*3),
+			httpclient.WithTimeout(time.Second*10, time.Second*60),
 		)
 	}
 	if b.i18nLoader == nil {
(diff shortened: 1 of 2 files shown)


bundle/bundle.go
```diff
@@ -46,7 +46,7 @@ func New(options ...Option) *Bundle {
 	}
 	if b.hc == nil {
 		b.hc = httpclient.New(
-			httpclient.WithTimeout(time.Second, time.Second*3),
+			httpclient.WithTimeout(time.Second*10, time.Second*60),
```
sfwn, line 49: Too long timeout config may lead 504 in the UI.
I think default 10s & 60s is enough.

Adjust in specific module's method is more recommended.

sfwn (changes requested):

bundle/bundle.go
johnlanni: The dialing timeout of 1s is too short. This change is for using telepresence by any modules.
10s for dial timeout & 60s for total timeout  are all reasonable.  UI nginx's read timeout is more than 60s, that's ok.

sfwn (approved):

PR #1040, opened 2021-07-16
pipeline pod namespace gc compensator

kakj-go:
#### What type of this PR
/kind bug

time end may be empty

### DIFF
--- a/modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
+++ b/modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
@@ -34,7 +34,7 @@ func (r *Reconciler) CompensateGCNamespaces() {
 	r.doWaitGCCompensate(true)
 
 	time.AfterFunc(24*time.Hour, func() {
-		go r.CompensateGCNamespaces()
+		r.CompensateGCNamespaces()
 	})
 }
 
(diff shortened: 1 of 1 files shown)


modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
```diff
@@ -89,7 +90,7 @@ func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pip
 				ttl = defaultGCTime
 			}
 
-			if p.TimeEnd.IsZero() {
+			if p.TimeEnd == nil || p.TimeEnd.IsZero() {
```
sfwn, line 93: if TimeEnd is empty and status is endStatus, you can use TimeUpdate field.
Otherwise, these namespaces will not be gc.

sfwn (changes requested):

modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
```diff
@@ -89,12 +90,14 @@ func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pip
 				ttl = defaultGCTime
 			}
 
-			if p.TimeEnd.IsZero() {
-				continue
-			}
-
-			if uint64(time.Now().Unix()-p.TimeEnd.Unix()) < (ttl + bufferTime) {
-				continue
+			if p.TimeEnd == nil {
```
sfwn, line 93: too ugly

sfwn: use a variable like timeEnd.
```go
timeEnd = p.timeEnd
if timeEnd == nil {
  timeEnd = p.timeUpdate
}
// logic xxx
```

sfwn (changes requested):

sfwn (approved):

PR #1048, opened 2021-07-16
fix orchestrator nil pointer dereference error

QvodSoldier:
#### What type of this PR

Add one of the following kinds:
/kind bug

#### What this PR does / why we need it:
fix deploy panic when addon config is empty

#### Which issue(s) this PR fixes:
- [nil point interface](https://erda.cloud/erda/workBench/projects/387/issues/all?id=74081&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDQzNiJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjF9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=-1&type=BUG)
- [personal mssage](https://erda.cloud/erda/workBench/projects/387/issues/all?id=73840&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDQzNiJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjF9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=429&type=BUG)
[...]

PR #1051, opened 2021-07-16
timing patch task executors and manual trigger refresh

chengjoey:
#### What type of this PR

/kind bug

task manager timing refresh executors sync
add endpoint to manual refresh executors

### DIFF
--- a/modules/pipeline/initialize.go
+++ b/modules/pipeline/initialize.go
@@ -134,10 +134,8 @@ func do() (*httpserver.Server, error) {
 		CronNotExecuteCompensate: pipelineSvc.CronNotExecuteCompensateById,
 	}
 
-	// init cluster info before pipeline scheduler task executors init
-	if err := clusterinfo.Initialize(bdl); err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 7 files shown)


sfwn:
please see the conversations of #1048

modules/pipeline/initialize.go
```diff
@@ -134,10 +134,7 @@ func do() (*httpserver.Server, error) {
 		CronNotExecuteCompensate: pipelineSvc.CronNotExecuteCompensateById,
 	}
 
-	// init cluster info before pipeline scheduler task executors init
-	if err := clusterinfo.Initialize(bdl); err != nil {
-		return nil, err
-	}
+	clusterinfo.Initialize(bdl)
```
sfwn, line 137: Why remove comments ?

sfwn: I think you still need to handle error at startup.

sfwn: Or you will have a `fake` healthy pipeline but with no executors to execute.

sfwn (changes requested):

sfwn (approved):

PR #1053, opened 2021-07-16
Fix/pipeline events flinkstatus

chengjoey:
#### What type of this PR

/kind bug

set status running if flinkstatus is empty
get projectName, appName from mergeLabels when send event


chengjoey:
If query status immediately after create flinkcluster, will get empty status, but actually it`s running

sfwn (approved):

PR #1081, opened 2021-07-19
Fix pipeline filetree node, add url unescape before decoding node path

shuofan:
#### What type of this PR
bug

Fix pipeline filetree node, add url unescape before decoding node path

#### Specified Reviewers:

/assign @sfwn

### DIFF
--- a/modules/dop/endpoints/projectpipelinefiletree.go
+++ b/modules/dop/endpoints/projectpipelinefiletree.go
@@ -70,7 +70,7 @@ func (e *Endpoints) GetFileTreeNode(ctx context.Context, r *http.Request, vars m
 	}
 	inode, err := url.QueryUnescape(vars["inode"])
 	if err != nil {
-		logrus.Error(apierrors.ErrGetFileTreeNode.InvalidParameter("QueryUnescape inode"))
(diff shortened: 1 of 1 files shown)


modules/dop/endpoints/projectpipelinefiletree.go
```diff
@@ -65,7 +68,13 @@ func (e *Endpoints) GetFileTreeNode(ctx context.Context, r *http.Request, vars m
 	if err := e.queryStringDecoder.Decode(&req, r.URL.Query()); err != nil {
 		return apierrors.ErrListFileTreeNodes.InvalidParameter(err).ToResp(), nil
 	}
-	req.Inode = vars["inode"]
+	inode, err := url.QueryUnescape(vars["inode"])
+	if err != nil {
+		logrus.Error(apierrors.ErrGetFileTreeNode.InvalidParameter("QueryUnescape inode"))
```
sfwn, line 73: use `logrus.Errorf` and print invalid inode

sfwn (changes requested):

modules/dop/endpoints/projectpipelinefiletree.go
shuofan: done

PR #1093, opened 2021-07-20
Add number of waiting tasks by types in test file records response

shuofan:
#### What type of this PR
feature

Add number of waiting tasks by types in test file records response

#### Specified Reviewers:

/assign @Effet @sfwn

### DIFF
--- a/apistructs/file_records.go
+++ b/apistructs/file_records.go
@@ -89,5 +89,10 @@ type GetTestFileRecordResponse struct {
 
 type ListTestFileRecordsResponse struct {
 	Header
-	Data []TestFileRecord
+	Data *ListTestFileRecordsResponseData
+}
+
+type ListTestFileRecordsResponseData struct {
+	Counter map[string]int   `json:"counter"`
+	List    []TestFileRecord `json:"list"`
 }
--- a/modules/dop/dao/file_records.go
+++ b/modules/dop/dao/file_records.go
@@ -84,6 +84,11 @@ func (client *DBClient) GetRecord(id uint64) (*TestFileRecord, error) {
 	return &res, nil
 }
 
+type stateCounter struct {
+	Type  string
+	Count int
+}
+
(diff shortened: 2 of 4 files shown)


apistructs/file_records.go
```diff
@@ -89,5 +89,10 @@ type GetTestFileRecordResponse struct {
 
 type ListTestFileRecordsResponse struct {
 	Header
-	Data []TestFileRecord
+	Data *ListTestFileRecordsResponseData
```
Effet, line 92: struct changes will break api compatibility, how do you deal with it?

modules/dop/dao/file_records.go
```diff
@@ -85,18 +85,28 @@ func (client *DBClient) GetRecord(id uint64) (*TestFileRecord, error) {
...
-	return res, nil
+
+	var count int
+	counter := make(map[string]int)
+	for _, t := range req.Types {
+		if err := client.Model(&TestFileRecord{}).Where("`type` = ? AND `state` = ?", t, apistructs.FileRecordStatePending).Count(&count).Error; err != nil {
+			return nil, nil, err
+		}
+		counter[string(t)] = count
+	}
```
Effet, line 107: for loop to query db is dangerous, you should use native sql `group by`.

Effet (changes requested):

modules/dop/dao/file_records.go
shuofan: done

apistructs/file_records.go
shuofan: It is not backward-compatible, as not designed well at the beginning. We may version it in the future. Updated API docs here: https://erda.cloud/erda/workBench/projects/387/apps/5880/apiDesign

Effet: so must frontend change their code at same time?

Effet (approved):

PR #1105, opened 2021-07-21
namespace gc count result num

kakj-go:
#### What type of this PR

/kind bug

namespace gc count result num not sql result total

### DIFF
--- a/modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
+++ b/modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
@@ -60,7 +60,7 @@ func (r *Reconciler) doWaitGCCompensate(isSnippetPipeline bool) {
 	}
 }
 
-func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pipeline, int64, error) {
+func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pipeline, int, error) {
 	var pipelineResults []spec.Pipeline
 
 	var req apistructs.PipelinePageListRequest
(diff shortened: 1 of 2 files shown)


modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
```diff
@@ -60,7 +60,7 @@ func (r *Reconciler) doWaitGCCompensate(isSnippetPipeline bool) {
 	}
 }
 
-func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pipeline, int64, error) {
+func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pipeline, int, error) {
```
Effet, line 63: explain why int64 -> int

modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
```diff
@@ -72,7 +72,7 @@ func (r *Reconciler) getNeedGCPipelines(pageNum int, isSnippet bool) ([]spec.Pip
 	}
 	req.AllSources = true
 	req.IncludeSnippet = isSnippet
-	pipelines, _, total, _, err := r.dbClient.PageListPipelines(req)
+	pipelines, _, _, _, err := r.dbClient.PageListPipelines(req)
```
Effet, line 75: PageListPipelines has too many returns, should refactor it

Effet (changes requested):

modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
kakj-go: result `len(pipelines)` was int type

modules/pipeline/pipengine/reconciler/pipeline_gc_compensator.go
kakj-go: yes, but not now

sfwn (approved):

PR #1108, opened 2021-07-21
use rollback config to do release gc

QvodSoldier:
#### What type of this PR
/kind bug

we need according to the rollback configuration to do release gc

- [Erda Cloud Issue Link](https://terminus-org.app.terminus.io/erda/workBench/projects/387/issues/all?id=70729&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDQzNiJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjF9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=431&type=TASK)

#### Specified Reviewers:

/assign @Effet

### DIFF
--- a/modules/core-services/services/project/project.go
+++ b/modules/core-services/services/project/project.go
@@ -437,7 +437,7 @@ func (p *Project) FillQuota(orgResources map[uint64]apistructs.OrgResourceInfo) 
 	return nil
 }
 
-// ListProjects
+// GetAllProjects list all project
 func (p *Project) GetAllProjects() ([]apistructs.ProjectDTO, error) {
 	projects, err := p.db.GetAllProjects()
 	if err != nil {
--- a/modules/core-services/services/project/project.go
+++ b/modules/core-services/services/project/project.go
@@ -438,8 +438,16 @@ func (p *Project) FillQuota(orgResources map[uint64]apistructs.OrgResourceInfo) 
 }
 
 // ListProjects
-func (p *Project) GetAllProjects() ([]model.Project, error) {
-	return p.db.GetAllProjects()
+func (p *Project) GetAllProjects() ([]apistructs.ProjectDTO, error) {
+	projects, err := p.db.GetAllProjects()
+	if err != nil {
+		return nil, err
+	}
+	projectsDTO := make([]apistructs.ProjectDTO, 0, len(projects))
+	for _, v := range projects {
+		projectsDTO = append(projectsDTO, p.convertToProjectDTO(true, &v))
+	}
+	return projectsDTO, nil
 }
 
 // ListAllProjects 企业管理员可查看当前企业下所有项目，包括未加入的项目
--- a/modules/orchestrator/services/runtime/runtime.go
+++ b/modules/orchestrator/services/runtime/runtime.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"runtime/debug"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/erda-project/erda/apistructs"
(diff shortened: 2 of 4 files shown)


modules/core-services/services/project/project.go
```diff
@@ -438,8 +438,16 @@ func (p *Project) FillQuota(orgResources map[uint64]apistructs.OrgResourceInfo)
 }
 
 // ListProjects
-func (p *Project) GetAllProjects() ([]model.Project, error) {
-	return p.db.GetAllProjects()
+func (p *Project) GetAllProjects() ([]apistructs.ProjectDTO, error) {
```
Effet, line 441: Comment and func name not match

modules/orchestrator/services/runtime/runtime.go
```diff
@@ -1698,11 +1704,11 @@ func (r *Runtime) FullGC() {
 			break
 		}
 		for i := range runtimes {
-			if runtimes[i].Workspace == string(apistructs.ProdWorkspace) {
-				r.fullGCForSingleRuntime(runtimes[i].ID, keep)
-			} else {
-				r.fullGCForSingleRuntime(runtimes[i].ID, 1)
+			keep, ok := rollbackCfg[runtimes[i].ProjectID][strings.ToUpper(runtimes[i].Workspace)]
```
Effet, line 1707: should check array boundary? `rollbackCfg[runtimes[i].ProjectID]` may nil

modules/orchestrator/services/runtime/runtime.go
```diff
@@ -1698,11 +1704,11 @@ func (r *Runtime) FullGC() {
...
+			if !ok {
+				keep = 5
 			}
```
Effet, line 1710: what if keep's value invalid? some like -1, 0 or much greater than 1000000

modules/orchestrator/services/runtime/runtime.go
```diff
@@ -1688,7 +1689,12 @@ func (r *Runtime) FullGC() {
 		}
 	}()
 
-	keep := 5
+	rollbackCfg, err := r.getRollbackConfig()
```
Effet, line 1692: it's high dangerous to query all projects in memory.

Effet (changes requested):

modules/orchestrator/services/runtime/runtime.go
QvodSoldier: I checked the actual data and there is no more than 300. This function will not calculate other data in the project, such as issue. So the risk is not high for now. 

The temporary cache solution feels more risky, we can do detailed cache solution later in the future.

modules/core-services/services/project/project.go
QvodSoldier: done

modules/orchestrator/services/runtime/runtime.go
QvodSoldier: `rollbackCfg` is also a map

modules/orchestrator/services/runtime/runtime.go
QvodSoldier: nice review, it will change to between 0 and 100. done

modules/orchestrator/services/runtime/runtime.go
```diff
@@ -1698,11 +1704,11 @@ func (r *Runtime) FullGC() {
 			break
 		}
 		for i := range runtimes {
-			if runtimes[i].Workspace == string(apistructs.ProdWorkspace) {
-				r.fullGCForSingleRuntime(runtimes[i].ID, keep)
-			} else {
-				r.fullGCForSingleRuntime(runtimes[i].ID, 1)
+			keep, ok := rollbackCfg[runtimes[i].ProjectID][strings.ToUpper(runtimes[i].Workspace)]
+			if !ok || keep < 0 || keep > 100 {
```
Effet, line 1708: keep <= 0 is correct

```suggestion
			if !ok || keep <= 0 || keep > 100 {
```

Effet (changes requested):

modules/orchestrator/services/runtime/runtime.go
Effet: please mark a comment TODO or FIXME, at least

modules/orchestrator/services/runtime/runtime.go
QvodSoldier: done

modules/orchestrator/services/runtime/runtime.go
QvodSoldier: done

Effet (approved):

PR #1118, opened 2021-07-21
Search test plan cases by unassigned executor

shuofan:
#### What type of this PR
feature

Search test plan cases by unassigned executor
#### Specified Reviewers:

/assign @Effet @sfwn

### DIFF
--- a/modules/dop/dao/testplan_testcase_relation.go
+++ b/modules/dop/dao/testplan_testcase_relation.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/erda-project/erda/apistructs"
 	"github.com/erda-project/erda/pkg/database/dbengine"
+	"github.com/erda-project/erda/pkg/strutil"
 )
 
 // TestPlanCaseRel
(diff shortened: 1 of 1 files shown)


modules/dop/dao/testplan_testcase_relation.go
```diff
@@ -161,7 +162,11 @@ func (client *DBClient) ListTestPlanCaseRels(req apistructs.TestPlanCaseRelListR
 		sql = sql.Where("`updater_id` IN (?)", req.UpdaterIDs)
 	}
 	if len(req.ExecutorIDs) > 0 {
-		sql = sql.Where("`executor_id` IN (?)", req.ExecutorIDs)
+		if strutil.Exist(req.ExecutorIDs, "unassigned") {
```
Effet, line 165: explain what unassigned means?

Effet (changes requested):

modules/dop/dao/testplan_testcase_relation.go
shuofan: Searching test cases which have not been assigned to any executor yet, use "unassigned" to distinguish empty executors([]string) cases in params, any suggestions？

sfwn (approved):

PR #1122, opened 2021-07-22
fix table audit_history error

QvodSoldier:
#### What type of this PR
/kind bug

If the fields of table `dice_audit_history` and table `dice_audit` are inconsistent, the audit archive will fail

#### Specified Reviewers:

/assign @Effet


Effet (changes requested):
since migration lint passed, can you explain how you maintain compatibility after change column from NULL to NOT NULL?

QvodSoldier:
The query sql used to be  `db.Where("fdp_project_id = ? or fdp_project_id is NULL", "")` , Should I delete ‘is null’ afterwards?
image

Effet:
`or` looks no problem for compaitibity

Effet (approved):

PR #1123, opened 2021-07-22
Url queryunescape inode in delete and find ancestor file tree node

shuofan:
#### What type of this PR
bug

before decode inode path, add url queryunescape inode in delete and find ancestor file tree node

#### Specified Reviewers:

/assign @sfwn

### DIFF
--- a/modules/dop/endpoints/projectpipelinefiletree.go
+++ b/modules/dop/endpoints/projectpipelinefiletree.go
@@ -159,8 +159,14 @@ func (e *Endpoints) DeleteFileTreeNode(ctx context.Context, r *http.Request, var
 		return apierrors.ErrDeleteFileTreeNode.NotLogin().ToResp(), nil
 	}
 
+	inode, err := url.QueryUnescape(vars["inode"])
+	if err != nil {
+		logrus.Errorf("QueryUnescape pipeline filetree inode %v failed", vars["inode"])
+		inode = vars["inode"]
+	}
+
 	req := apistructs.UnifiedFileTreeNodeDeleteRequest{
-		Inode:        vars["inode"],
+		Inode:        inode,
 		IdentityInfo: identityInfo,
 	}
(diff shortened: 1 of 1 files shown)


modules/dop/endpoints/projectpipelinefiletree.go
```diff
@@ -159,8 +159,14 @@ func (e *Endpoints) DeleteFileTreeNode(ctx context.Context, r *http.Request, var
 		return apierrors.ErrDeleteFileTreeNode.NotLogin().ToResp(), nil
 	}
 
+	inode, err := url.QueryUnescape(vars["inode"])
```
Effet, line 162: what purpose must unescape inode?

modules/dop/endpoints/projectpipelinefiletree.go
```diff
@@ -159,8 +159,14 @@ func (e *Endpoints) DeleteFileTreeNode(ctx context.Context, r *http.Request, var
...
+	if err != nil {
+		logrus.Errorf("QueryUnescape pipeline filetree inode %v failed", vars["inode"])
```
Effet, line 164: if err not returned, Warnf is better. And why not return err, and fallback to plain var["inode"] ?

Effet (changes requested):

modules/dop/endpoints/projectpipelinefiletree.go
shuofan: base64 stdEncoding params in url containing "/" result in 404, frontend will pass url encoding inode.

modules/dop/endpoints/projectpipelinefiletree.go
shuofan: if it is invalid URL escaper, treat it as ordinary inode

sfwn (approved):

PR #1171, opened 2021-07-27
update dicehub image addr

ZLValien:
#### What type of this PR

/kind feature

image service have been delete in endpoints. 
provider open 8080 port

### DIFF
--- a/erda.yml
+++ b/erda.yml
@@ -249,6 +249,9 @@ services:
       - port: 10000
         protocol: "TCP"
         l4_protocol: "TCP"
+      - port: 8080
+        protocol: "TCP"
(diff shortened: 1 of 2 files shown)


erda.yml
```diff
@@ -249,6 +249,9 @@ services:
       - port: 10000
         protocol: "TCP"
         l4_protocol: "TCP"
+      - port: 8080
```
sfwn, line 252: Does image provider have invoker?

And you should add a global env `DICEHUB_ADDR` point to original addr for discover random-select issue.

sfwn (changes requested):

PR #1182, opened 2021-07-28
doc: add v1.1 changeLog

daskyrk:
#### What type of this PR
/kind document

* Add v1.1 changeLog
* Update changeLog files directory


sfwn:
lgtm.

Imagine a question:
- we need a fix for v1.1.3, so the version is v1.1.4
- And latest v1.2.1 is already released. 

Where should we put the v1.1.4 into? Upper to latest v1.2.1 or between v1.2.0 and v1.1.3.

daskyrk:
> lgtm.
> 
> Imagine a question:
> 
> * we need a fix for v1.1.3, so the version is v1.1.4
> * And latest v1.2.1 is already released.
> 
> Where should we put the v1.1.4 into? Upper to latest v1.2.1 or between v1.2.0 and v1.1.3.

Good question, i think we should insert patch version into the ordered list. For better view the patch version, i add a table.

PR #1217, opened 2021-07-29
pipeline events show order

kakj-go:
#### What type of this PR

/kind bug

The order of the records of the pipeline events is modified. The original count number and the order of the records is not correct, events should be able to express the order instead of just counting.

erda-issue: [erda-issue](https://erda.cloud/erda/dop/projects/387/issues/all?id=200490&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDU2MCJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=506&type=BUG)

### DIFF
--- a/modules/pipeline/dbclient/op_pipeline_event.go
+++ b/modules/pipeline/dbclient/op_pipeline_event.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/erda-project/erda/apistructs"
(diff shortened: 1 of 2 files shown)


modules/pipeline/dbclient/op_pipeline_event.go
```diff
@@ -129,10 +94,78 @@ func (client *Client) AppendPipelineEvent(pipelineID uint64, newEvents []*apistr
...
+			ordered = append(ordered, ev)
+			lastEvent = ev
+		}
+	}
+	return ordered
+}
+
 func (o orderedEvents) Len() int           { return len(o) }
-func (o orderedEvents) Less(i, j int) bool { return o[i].FirstTimestamp.Before(o[j].FirstTimestamp) }
+func (o orderedEvents) Less(i, j int) bool { return o[i].LastTimestamp.Before(o[j].LastTimestamp) }
```
sfwn, line 168: Why use LastTimestamp, does it influence merge logic ?

sfwn (changes requested):

modules/pipeline/dbclient/op_pipeline_event.go
kakj-go: FirstTimestamp sorting affects aggregation，FirstTimestamp will get smaller and smaller

```golang
if !ev.FirstTimestamp.IsZero() && ev.FirstTimestamp.Before(lastEvent.FirstTimestamp) {
    lastEvent.FirstTimestamp = ev.FirstTimestamp
}
 if ev.LastTimestamp.After(lastEvent.LastTimestamp) {
    lastEvent.LastTimestamp = ev.LastTimestamp
}
```

sfwn (approved):

PR #1345, opened 2021-08-06
Fixed the problem of extension provider reading configuration failure and incorrect interface parameter parsing

ZLValien:
#### What type of this PR

/kind feature

extension provider is in module dicehub

.yml cannot read env like map[string][]string.So read it as string and unmarshal  it when provider init.

Many interface parameters were originally processed by url.QueryUnescape(), but they were omitted during refac. Now add them


sfwn:
Add labels.

Effet:
add unit test for real menu data

ZLValien:
> add unit test for real menu data

ZLValien:
> add unit test for real menu data

done 。add unit test for real data

PR #1378, opened 2021-08-10
concurrent query issue for workbench endpoint optimization 

chengjoey:
#### What type of this PR

/kind bug

concurrent query issue by expire day
optimization for workbench endpoint time consuming

### DIFF
--- a/modules/dop/dao/issue.go
+++ b/modules/dop/dao/issue.go
@@ -518,3 +518,104 @@ func (client *DBClient) GetReceiversByIssueID(issueID int64) ([]string, error) {
 
 	return strutil.DedupSlice(receivers), nil
 }
+
+// GetIssueNumByPros query by IssuePagingRequest and group by project_id to get special issue num
+func (client *DBClient) GetIssueNumByPros(projectIDS []uint64, req apistructs.IssuePagingRequest) ([]apistructs.IssueNum, error) {
+	var (
+		res []apistructs.IssueNum
+	)
+
+	sql := client.Table("dice_issues").Select("count(id) as issue_num, project_id").Where("deleted = ?", 0)
+	if len(req.IDs) > 0 {
+		sql = sql.Where("dice_issues.id IN (?)", req.IDs)
+	}
+	if len(req.IterationIDs) > 0 {
+		sql = sql.Where("iteration_id in (?)", req.IterationIDs)
+	}
+	if len(req.Type) > 0 {
+		sql = sql.Where("type IN (?)", req.Type)
+	}
+	if len(req.Creators) > 0 {
+		sql = sql.Where("creator IN (?)", req.Creators)
+	}
+	if len(req.Assignees) > 0 {
+		sql = sql.Where("assignee IN (?)", req.Assignees)
+	}
+	if len(req.Priority) > 0 {
+		sql = sql.Where("priority IN (?)", req.Priority)
+	}
+	if len(req.Complexity) > 0 {
+		sql = sql.Where("complexity IN (?)", req.Complexity)
+	}
+	if len(req.Severity) > 0 {
+		sql = sql.Where("severity IN (?)", req.Severity)
+	}
+	if len(req.State) > 0 {
+		sql = sql.Where("state IN (?)", req.State)
+	}
+	if len(req.Owner) > 0 {
+		sql = sql.Where("owner IN (?)", req.Owner)
+	}
+	if len(req.BugStage) > 0 {
+		sql = sql.Where("stage IN (?)", req.BugStage)
+	} else if len(req.TaskType) > 0 {
+		sql = sql.Where("stage IN (?)", req.TaskType)
+	}
+	if len(req.ExceptIDs) > 0 {
+		sql = sql.Not("id", req.ExceptIDs)
+	}
+	if req.StartCreatedAt > 0 {
+		startCreatedAt := time.Unix(req.StartCreatedAt/1000, 0)
+		sql = sql.Where("dice_issues.created_at >= ?", startCreatedAt)
+	}
+	if req.EndCreatedAt > 0 {
+		endCreatedAt := time.Unix(req.EndCreatedAt/1000, 0)
+		sql = sql.Where("dice_issues.created_at <= ?", endCreatedAt)
+	}
+	if req.IsEmptyPlanFinishedAt {
+		sql = sql.Where("plan_finished_at IS NULL")
+	}
+	if req.StartFinishedAt > 0 && !req.IsEmptyPlanFinishedAt {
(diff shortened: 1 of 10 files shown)


shuofan (approved):

modules/dop/endpoints/workbench.go
```diff
@@ -130,49 +133,56 @@ func (e *Endpoints) GetWorkbenchData(ctx context.Context, r *http.Request, vars
...
-				if timeList[index][1] != 0 {
-					etIssueReq.EndFinishedAt = timeList[index][1] * 1000
+			go func(idx int, ed string) {
+				defer wg.Done()
+				etIssueReq := apistructs.IssuePagingRequest{}
+				etIssueReq.OrgID = int64(workReq.OrgID)
+				etIssueReq.StartFinishedAt = timeList[idx][0] * 1000
+				if et == "unspecified" {
+					etIssueReq.IsEmptyPlanFinishedAt = true
+				} else {
```
Effet, line 146: it's better package this go func() as a real function, and give some unit test

Effet (changes requested):

Effet:
your pr title is not good. it only tells us how you changed code, but no mention about what you fixed.

modules/dop/endpoints/workbench.go
chengjoey: it`s hard to package this go func(), because Contextual connection

modules/dop/endpoints/workbench.go
```diff
@@ -77,102 +66,14 @@ func (e *Endpoints) GetWorkbenchData(ctx context.Context, r *http.Request, vars
...
-				Type: []apistructs.IssueType{
-					apistructs.IssueTypeRequirement,
-					apistructs.IssueTypeBug,
-					apistructs.IssueTypeTask,
-				},
-				Asc: true,
-			},
-		}
-		issues, total, err := e.issue.Paging(issueReq)
+		issueItem, err := e.workBench.GetUndoneProjectItem(userID.String(), workReq.IssueSize, v)
```
Effet, line 73: let me describe by pseudo-code, what you do now:

```
for v := range projects {
  for t := range type of issue state { // 7 types
      // this for is concurrently, but only calculate one project
  }
}
```
at the result, if there are 10 projects, it will have 10 x 7 = 70 times sql, as it run as concurrently, in user experience, it will have 10 times slow.

And what you can improve, I suggest you should:
```
for t := range type of issue state {
   bulk query and sum all projects (10 projects), only for this type
}
```
and this `for` can be concurrently, so as the result, there will be 7 sql no matter how many projects. In user experience, always fast like one sql.

Effet (changes requested):

modules/dop/endpoints/workbench.go
chengjoey: can`t query issue num by project list, becase query issue num should stateID list first, different project`s state list are different

modules/dop/dao/issue.go
```diff
@@ -518,3 +518,104 @@ func (client *DBClient) GetReceiversByIssueID(issueID int64) ([]string, error) {
 
 	return strutil.DedupSlice(receivers), nil
 }
+
+// GetSpecialIssueNum query by IssuePagingRequest and group by project_id to get special issue num
```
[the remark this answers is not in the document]
chengjoey: different plan finished time issues

modules/dop/services/workbench/workbench.go
```diff
@@ -0,0 +1,194 @@
...
+			} else {
+				if timeList[idx][1] != 0 {
+					etIssueReq.EndFinishedAt = timeList[idx][1] * 1000
+				}
+			}
+			etIssueReq.StateBelongs = StateBelongs
+			etIssueReq.External = true
+			etIssueReq.Type = IssueTypes
+			etIssueReq.Assignees = req.Assignees
+			prosIssueNumList, err := w.issueSvc.GetSpecialIssueNum(projectIDS, etIssueReq)
```
Effet, line 160: move the selected code into a new func

Effet (changes requested):

modules/dop/dao/issue.go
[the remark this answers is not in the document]
Effet: another good name?

modules/dop/services/workbench/workbench.go
chengjoey: the new func is simalar with w.issueSvc.GetSpecialIssueNum(), and new func need to custom timeList and projectIDS again

Effet: ok, your opinion makes sense, so be it.

modules/dop/dao/issue.go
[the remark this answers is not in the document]
Effet: @chengjoey you can change func name as suggested by @shuofan

modules/dop/endpoints/workbench.go
```diff
@@ -77,104 +66,18 @@ func (e *Endpoints) GetWorkbenchData(ctx context.Context, r *http.Request, vars
...
```
Effet, line 72: still one project one sql?

Effet (changes requested):

modules/dop/services/workbench/workbench.go
```diff
@@ -0,0 +1,194 @@
...
+
+// SetSpecialIssueNum concurrent query different expire issue num
+func (w *Workbench) SetSpecialIssueNum(req apistructs.IssuePagingRequest, items []*apistructs.WorkbenchProjectItem) error {
+	if len(items) == 0 {
+		return nil
+	}
+	var projectIDS []uint64
+	for _, item := range items {
+		projectIDS = append(projectIDS, item.ProjectDTO.ID)
+	}
```
Effet, line 124: if items only to used to retrieve projectIDs, you should let projectIDs as func param directly. you just wasted query too many sql to get the items, and I commented in another code block.

Effet (dismissed):

modules/dop/services/workbench/workbench.go
chengjoey: e.workBench.GetUndoneProjectItem get issue table by issue size, not only issue num, it`s hard to group by project id list.

modules/dop/dao/issue.go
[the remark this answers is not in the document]
chengjoey: changed func name to GetIssueNumsByPros, and another func to SetDiffFinishedIssueNum

modules/dop/endpoints/workbench.go
chengjoey: now still one project one sql, @shuofan may make optimization later

sfwn (approved):

PR #1488, opened 2021-08-17
fix : create extension failed when init provider

ZLValien:
#### What type of this PR

/kind bug

The read file address does not match the actual address

### DIFF
--- a/modules/dicehub/extension/provider.go
+++ b/modules/dicehub/extension/provider.go
@@ -34,6 +34,8 @@ type config struct {
 	ExtensionMenu map[string][]string `file:"extension_menu" env:"EXTENSION_MENU"`
 }
 
+const FilePath = "/app/extensions-init"
+
 // +provider
 type provider struct {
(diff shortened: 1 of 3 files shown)


modules/dicehub/extension/provider.go
```diff
@@ -63,7 +63,7 @@ func (p *provider) Init(ctx servicehub.Context) error {
 			))
 	}
 	go func() {
-		err := p.extensionService.InitExtension("/app/extension")
+		err := p.extensionService.InitExtension("/app/dicehub-extension")
```
sfwn, line 66: extensions

sfwn (changes requested):

modules/dicehub/extension/provider.go
ZLValien: done

sfwn (approved):

sfwn (approved):

PR #1510, opened 2021-08-19
when creating a pipeline to cache action information, add a lock to read the map

kakj-go:
#### What type of this PR

/kind bug

The passedDataWhenCreate method will read the map concurrently, which will cause the pipeline to restart

erda-issue: [erda-issue](https://terminus-org.app.terminus.io/erda/dop/projects/387/issues/all?id=206764&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDU2MCJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=431&type=BUG)

### DIFF
--- a/modules/pipeline/services/pipelinesvc/passedDataWhenCreate.go
+++ b/modules/pipeline/services/pipelinesvc/passedDataWhenCreate.go
@@ -27,9 +27,8 @@ import (
 // passedDataWhenCreate stores data passed recursively when create graph.
 type passedDataWhenCreate struct {
 	extMarketSvc     *extmarketsvc.ExtMarketSvc
-	actionJobDefines map[string]*diceyml.Job
-	actionJobSpecs   map[string]*apistructs.ActionSpec
-	lock             sync.Mutex
+	actionJobDefines *sync.Map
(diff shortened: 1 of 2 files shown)


sfwn:
Test if have dead lock.

kakj-go:
These methods are locked, but these methods do not call each other, they will be unlocked after the method ends, which does not meet the 4 elements of deadlock

sfwn (changes requested):
If no environment can be used to test, you should test it locally at least.

sfwn (approved):

PR #1599, opened 2021-08-30
add role auditor

QvodSoldier:
#### What type of this PR

/kind feature

add a auditor role

- Fixes #your-issue_number
- [Erda Cloud Issue Link](https://terminus-org.app.terminus.io/erda/dop/projects/387/issues/all?id=208355&issueFilter__urlQuery=eyJpdGVyYXRpb25JRHMiOls1MDZdLCJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDQzNiJdfQ%3D%3D&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=506&type=TASK)

#### Specified Reviewers:

/assign @Effet


modules/core-services/endpoints/member.go
```diff
@@ -148,7 +148,8 @@ func (e *Endpoints) ListMemberRoles(ctx context.Context, r *http.Request, vars m
 			}
 		}
 	}
-	if orgID == 0 {
+
+	if orgID == 0 && scopeType != apistructs.SysScope {
```
Effet, line 152: different condition should return different error

modules/core-services/services/member/member.go
```diff
@@ -602,7 +602,18 @@ func (m *Member) CheckPermission(userID string, scopeType apistructs.ScopeType,
...
+	if err != nil {
+		return err
+	}
+
 	switch scopeType {
+	case apistructs.SysScope:
+		// 系统管理员
+		if admin {
+			return nil
+		}
```
Effet, line 616: why not check admin user's permission as same as other user?

I think you just query `m.db.GetMemberByScopeAndUserID(userID, scopeType, scopeID)` also work for admin?

thus `m.db.IsSysAdmin` may useless?

modules/dop/endpoints/member.go
```diff
@@ -60,7 +60,8 @@ func (e *Endpoints) ListMemberRoles(ctx context.Context, r *http.Request, vars m
 			}
 		}
 	}
-	if orgID == 0 {
+
+	if orgID == 0 && scopeType != apistructs.SysScope {
```
Effet, line 64: different condition should return different error

Effet (changes requested):

modules/core-services/endpoints/member.go
QvodSoldier: I think it's the same condition that means the orgID can't be equal to 0, only if the scopeType is 'sys'.

modules/dop/endpoints/member.go
QvodSoldier: I think it's the same condition that means the orgID can't be equal to 0, only if the scopeType is 'sys'.

modules/core-services/services/member/member.go
QvodSoldier: `m.db.GetMemberByScopeAndUserID(userID, scopeType, scopeID)` will return all the members in this scope, and if you use this function, you need to do another traversal outside.  I think this old function is also quite good.

sfwn:
add unit test to match coverage requirement.

Effet (approved):

PR #1610, opened 2021-08-31
magic temp hack: dop inject DOP_ADDR to localhost:9527 to fix internal invoke timeout issue in cp

sfwn:
#### What type of this PR

/kind bugfix

magic temp hack: dop inject DOP_ADDR to localhost:9527 to fix internal invoke timeout issue in cp.

TODO change to internal method invoke in component-protocol and remove this hack.

#### Specified Reviewers:

/assign @Effet

### DIFF
--- a/erda.yml
+++ b/erda.yml
@@ -58,7 +58,7 @@ services:
       replicas: 1
     envs:
       DEBUG: "true"
-      MONITOR_ADDR: monitor:7096
+      DOP_ADDR: localhost:9527 # TODO remove it after internal bundle invoke inside cp issue-manage adjusted
     health_check:
       exec: { }
       http:


erda.yml
```diff
@@ -58,7 +58,7 @@ services:
       replicas: 1
     envs:
       DEBUG: "true"
-      MONITOR_ADDR: monitor:7096
```
Effet, line 61: MONITOR_ADDR useless?

Effet (changes requested):

erda.yml
sfwn: already defined globally in erda.yml global envs.

sfwn: see https://github.com/erda-project/erda/blob/master/erda.yml#L13

Effet (approved):

PR #1630, opened 2021-09-01
pipeline remove default cron compensator

kakj-go:
#### What type of this PR
/kind bug

The default timing compensation will make users think that this is a timing task switch, so the default timing compensation configuration is removed

erda-issue: [erda-issue](https://erda.cloud/erda/dop/projects/387/issues/all?id=210303&issueFilter__urlQuery=eyJpdGVyYXRpb25JRHMiOls1MDZdLCJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDU2MCJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=506&type=TASK)

### DIFF
--- a/pkg/parser/pipelineyml/visitor_cron.go
+++ b/pkg/parser/pipelineyml/visitor_cron.go
@@ -120,11 +120,6 @@ func (v *CronVisitor) Visit(s *Spec) {
 		v.nextTimes = append(v.nextTimes, nextTime)
 		scheduleFrom = nextTime
 	}
-
-	// compensator
-	if s.CronCompensator == nil {
-		s.CronCompensator = &DefaultCronCompensator
-	}
 }
 
 func ListNextCronTime(cronExpr string, ops ...CronVisitorOption) ([]time.Time, error) {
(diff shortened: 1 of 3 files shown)


pkg/parser/pipelineyml/visitor_cron.go
```diff
@@ -120,11 +120,6 @@ func (v *CronVisitor) Visit(s *Spec) {
 		v.nextTimes = append(v.nextTimes, nextTime)
 		scheduleFrom = nextTime
 	}
-
-	// compensator
-	if s.CronCompensator == nil {
-		s.CronCompensator = &DefaultCronCompensator
-	}
```
sfwn, line 127: Can you delete yaml part already with default con compensator?
Users don't care about this. And they can get polished default compensator configs if we adjusted.

sfwn (changes requested):

sfwn (approved):

PR #1642, opened 2021-09-01
database gc When the ttl time arrives, you should not care whether it is ns gc

kakj-go:
#### What type of this PR
/kind bug

Compensation for database gc, some data may not be ns gc, an error will be reported when calling the interface

erda-issue: [erda-issue](https://erda.cloud/erda/dop/projects/387/issues/all?id=72979&issueFilter__urlQuery=eyJpdGVyYXRpb25JRHMiOls1MDZdLCJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDU2MCJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=506&type=BUG)

### DIFF
--- a/modules/pipeline/dbclient/op_pipeline_delete.go
+++ b/modules/pipeline/dbclient/op_pipeline_delete.go
@@ -16,6 +16,7 @@ package dbclient
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/erda-project/erda/modules/pipeline/spec"
 )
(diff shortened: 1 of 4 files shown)


modules/pipeline/dbclient/op_pipeline_delete.go
```diff
@@ -67,6 +68,12 @@ func canDelete(p spec.Pipeline) (bool, string) {
 	}
 	// 终态后需要判断 complete gc
 	if p.Status.IsEndStatus() {
+		if p.TimeCreated != nil && p.Extra.GC.DatabaseGC.Finished.TTLSecond != nil {
```
sfwn, line 71: Use timeEnd(use timeUpdated if end is empty), may be it is a log-run job.

modules/pipeline/dbclient/op_pipeline_delete.go
```diff
@@ -67,6 +68,12 @@ func canDelete(p spec.Pipeline) (bool, string) {
...
+			ok := p.TimeCreated.Before(time.Now().Add(time.Duration(int64(*p.Extra.GC.DatabaseGC.Finished.TTLSecond)) * time.Second * -1))
```
sfwn, line 72: .Add(-time.Duration....)

sfwn (changes requested):

PR #1683, opened 2021-09-06
Workbench data and issue query optimization by issue expiry status

shuofan:
#### What type of this PR
feature

Workbench data and issue query optimization by issue expiry status

#### Specified Reviewers:

/assign @Effet @sfwn

### DIFF
--- a/modules/dop/dao/issue.go
+++ b/modules/dop/dao/issue.go
@@ -685,22 +685,22 @@ func (client *DBClient) GetIssuesByProject(req apistructs.IssuePagingRequest) ([
 	return res, total, nil
 }
 
+var conditions = map[ExpireType]string{
+	ExpireTypeUndefined:      "a.plan_finished_at IS NULL",
+	ExpireTypeExpired:        "a.plan_finished_at < CURDATE()",
+	ExpireTypeExpireIn1Day:   "a.plan_finished_at = CURDATE()",
+	ExpireTypeExpireIn2Days:  "a.plan_finished_at = DATE_ADD(CURDATE(),INTERVAL 1 DAY)",
+	ExpireTypeExpireIn7Days:  "a.plan_finished_at > DATE_ADD(CURDATE(),INTERVAL 1 DAY) AND a.plan_finished_at < DATE_ADD(CURDATE(),INTERVAL 7 DAY)",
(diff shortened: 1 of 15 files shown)


modules/dop/services/workbench/workbench.go
```diff
@@ -95,7 +88,7 @@ func (w *Workbench) GetUndoneProjectItem(userID string, issueSize int, pro apist
 		PageSize: uint64(issueSize),
 		IssueListRequest: apistructs.IssueListRequest{
 			ProjectID:    uint64(pro.ID),
-			StateBelongs: StateBelongs,
+			StateBelongs: apistructs.StateBelongs,
```
Effet, line 91: func `GetUndoneProjectItem` should be deleted?

modules/dop/dao/issue.go
```diff
@@ -620,3 +643,64 @@ func (client *DBClient) GetIssueNumByPros(projectIDS []uint64, req apistructs.Is
...
+	  WHEN a.plan_finished_at < CURDATE() THEN 'Expired'
+	  WHEN a.plan_finished_at = CURDATE() THEN 'ExpireIn1Day'
+	  WHEN a.plan_finished_at = DATE_ADD(CURDATE(),INTERVAL 1 DAY) THEN 'ExpireIn2Day'
+	  WHEN a.plan_finished_at < DATE_ADD(CURDATE(),INTERVAL 7 DAY) THEN 'ExpireIn7Day'
+	  WHEN a.plan_finished_at < DATE_ADD(CURDATE(),INTERVAL 30 DAY) THEN 'ExpireIn30Day'
+	  ELSE 'ExpireInFuture'
+	END
+	)
+    WHERE a.plan_finished_at IS NOT NULL AND b.belong IN (?)
+	`
```
Effet, line 704: I think you should split to 7 sql, because your current sql may update full-table.

```sql
UPDATE dice_issues a LEFT JOIN dice_issue_state b ON a.state = b.id
SET a.expiry_status = 'Expired'
WHERE a.plan_finished_at < CURDATE() AND a.expiry_status != 'Expired'
```
and so on, the key is **a.expiry_status != 'Expired'**.

Effet (changes requested):

modules/dop/services/workbench/workbench.go
shuofan: done

modules/dop/dao/issue.go
shuofan: After we split sql and add index. For single sql with one million data, time reduced from 13s to 1.7s

Effet (approved):

PR #1705, opened 2021-09-07
fix BUG: duplicate data in the test plan list and the env of plan is not refreshed

ZLValien:
#### What type of this PR

/kind bug

func (tp *TestPlanV2Join) Convert2DTO() will cause duplicate executeTime
change  *TestPlanV2Join to TestPlanV2Join

plan env should refresh when click ExecuteHistoryTable

### DIFF
--- a/modules/openapi/component-protocol/scenarios/auto-test-plan-detail/protocol.yml
+++ b/modules/openapi/component-protocol/scenarios/auto-test-plan-detail/protocol.yml
@@ -320,6 +320,38 @@ rendering:
       state:
         - name: "pipelineId"
           value: "{{ executeHistoryTable.pipelineId }}"
+    - name: envDrawer
+    - name: envBaseInfoTitle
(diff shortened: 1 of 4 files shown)


modules/openapi/component-protocol/scenarios/auto-test-plan-detail/protocol.yml
```diff
@@ -320,6 +320,38 @@ rendering:
       state:
         - name: "pipelineId"
           value: "{{ executeHistoryTable.pipelineId }}"
+    - name: envDrawer
+    - name: envBaseInfoTitle
+    - name: envBaseInfo
+      state:
+        - name: "envData"
+          value: "{{ executeInfo.envData }}"
```
sfwn, line 328: maybe put in global state?

sfwn (changes requested):

sfwn (approved):

PR #1727, opened 2021-09-08
pipeline legacy and infra http server use the same route and port

kakj-go:
#### What type of this PR
/kind feature

When protobuf adds an open interface, the address port of the route will be old, and it is impossible to specify a route to any port, which causes the open interface declared by protobuf to be unable to be routed, so the pipeline components need to be changed to use the same routing port

### DIFF
--- a/conf/pipeline/pipeline.yaml
+++ b/conf/pipeline/pipeline.yaml
@@ -3,7 +3,7 @@ pipeline:
 service-register:
 
 http-server:
-  addr: ":3082"
+  addr: ":3081"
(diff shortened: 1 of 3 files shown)


sfwn:
Add your tested screenshot plz.

sfwn:
These changes have huge effect.

kakj-go:
> Add your tested screenshot plz.

## The result of running the pipe on my computer

image

## curl api result

image

## web page visit result
image

sfwn (approved):

PR #1728, opened 2021-09-08
feature: org sync nexus configs only if crossCluster or isPublisher

sfwn:
#### What type of this PR

/kind feature

Org-level nexus configs is heavy. Most org doesn't need these nexus configs.
If no nexus info, always return false.
Otherwise, judge by crossCluster and isPublisher.

#### Specified Reviewers:

/assign @Effet


sfwn:
Test:
- create org and no ingress generated
- update org enable release-cross-cluster
- ingress generated
image

Effet (approved):

PR #1781, opened 2021-09-10
automated test step execution optimization

kakj-go:
#### What type of this PR
/kind feature

Automated test step running uses the last execution record of the current scene to render the ${{ outputs.xxx.xxx }} placeholders in it

##### run scene pre step output
image

##### run step use pre step value

image

##### run step rendered request params value 
image

### DIFF
--- a/modules/dop/services/autotest_v2/scene.go
+++ b/modules/dop/services/autotest_v2/scene.go
@@ -525,6 +525,8 @@ func (svc *Service) ExecuteDiceAutotestSceneStep(req apistructs.AutotestExecuteS
 		apiTestStr = strings.ReplaceAll(apiTestStr, expression.OldLeftPlaceholder+expression.Params+"."+param.Name+expression.OldRightPlaceholder, expression.ReplaceRandomParams(param.Temp))
 	}
 
+	apiTestStr = svc.renderPreSceneStepsOutput(step.SceneID, apiTestStr)
+
 	for _, conf := range configs.Data {
 		switch conf.Key {
 		case autotest.CmsCfgKeyAPIGlobalConfig:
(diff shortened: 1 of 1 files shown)


modules/dop/services/autotest_v2/scene.go
```diff
@@ -525,6 +525,32 @@ func (svc *Service) ExecuteDiceAutotestSceneStep(req apistructs.AutotestExecuteS
 		apiTestStr = strings.ReplaceAll(apiTestStr, expression.OldLeftPlaceholder+expression.Params+"."+param.Name+expression.OldRightPlaceholder, expression.ReplaceRandomParams(param.Temp))
 	}
 
+	var pipelinePageListRequest = apistructs.PipelinePageListRequest{
+		PageNum:  1,
+		PageSize: 10,
```
sfwn, line 530: why get 10 but only use one (index 0)?

sfwn (changes requested):

modules/dop/services/autotest_v2/scene.go
```diff
@@ -525,6 +525,32 @@ func (svc *Service) ExecuteDiceAutotestSceneStep(req apistructs.AutotestExecuteS
...
+		if err == nil && pipeline != nil {
+			for _, stage := range pipeline.PipelineStages {
+				for _, task := range stage.PipelineTasks {
+					for _, result := range task.Result.Metadata {
+						apiTestStr = strings.ReplaceAll(apiTestStr, expression.GenOutputRef(task.Name, result.Name), result.Value)
+					}
+				}
+			}
+		}
+	}
```
Effet, line 552: too deep, split to a func or multiple funcs

modules/dop/services/autotest_v2/scene.go
kakj-go: Really shouldn't query 10

Get the subscript 0, just to get the latest end record，I added sorting by id should be no problem

Effet (dismissed):

sfwn (approved):

PR #1824, opened 2021-09-13
detail return pipeline disable action

kakj-go:
#### What type of this PR
/kind bug

detail return pipeline disable action

####  test image

image


sfwn (approved):

sfwn:
you need add unit-test.

sfwn (approved):

PR #1837, opened 2021-09-14
multi-instance dicehub synchronizes local extension

kakj-go:
#### What type of this PR
/kind bug

When multi instance dicehub synchronizes the extension, it is set that only a single instance can be synchronized

erda-issue: [erda-issue](https://erda-org.erda.cloud/erda/dop/projects/387/issues/all?id=210670&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXSwiYXNzaWduZWVJRHMiOlsiMTAwMDU2MCJdfQ%3D%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6ImthbmJhbiIsImNoaWxkcmVuVmFsdWUiOnsia2FuYmFuIjoiZGVhZGxpbmUifX0%3D&iterationID=506&type=BUG)

test image:
image

image

### DIFF
--- a/modules/dicehub/extension/provider.go
+++ b/modules/dicehub/extension/provider.go
@@ -15,7 +15,9 @@
 package extension
 
 import (
+	"context"
 	"net/http"
+	"sync"
 
 	"github.com/jinzhu/gorm"
 	"github.com/robfig/cron"
@@ -26,6 +28,7 @@ import (
 	transport "github.com/erda-project/erda-infra/pkg/transport"
 	transhttp "github.com/erda-project/erda-infra/pkg/transport/http"
 	"github.com/erda-project/erda-infra/pkg/transport/http/encoding"
+	election "github.com/erda-project/erda-infra/providers/etcd-election"
 	pb "github.com/erda-project/erda-proto-go/core/dicehub/extension/pb"
 	"github.com/erda-project/erda/bundle"
 	"github.com/erda-project/erda/modules/dicehub/extension/db"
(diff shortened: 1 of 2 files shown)


modules/dicehub/extension/provider.go
```diff
@@ -63,13 +69,29 @@ func (p *provider) Init(ctx servicehub.Context) error {
...
-		if err != nil {
-			panic(err)
-		}
-		logrus.Infoln("End init extension")
-	}()
+
+	var once sync.Once
+	p.InitExtensionElection.OnLeader(func(ctx context.Context) {
+		once.Do(func() {
+			var doneValue atomic.Value
```
Effet, line 76: once already use atomic value to prevent concurrency invoke, is this `doneValue` needless?

cc @sfwn

Effet (dismissed):

modules/dicehub/extension/provider.go
sfwn: no need.

modules/dicehub/extension/provider.go
```diff
@@ -63,13 +69,29 @@ func (p *provider) Init(ctx servicehub.Context) error {
...
+			}
+			doneValue.Store(true)
+			logrus.Infoln("End init extension")
+
+			go func() {
+				select {
+				case <-ctx.Done():
+					done, ok := doneValue.Load().(bool)
+					if !ok || !done {
+						panic(fmt.Errorf("init extension not done but lost leader"))
```
sfwn, line 89: No need panic if leader lost.
Another leader will initExtensions.

And your go func should put before line #77.

sfwn (changes requested):

sfwn (approved):

PR #1861, opened 2021-09-14
autotest plan action show search input, and set pageSize

kakj-go:
#### What type of this PR
/kind bug

The data queried in the action drop-down box executed by the automated test plan is missing. All data should be queried, and then the search box should be searched.

#### test image
image

image

### DIFF
--- a/modules/openapi/component-protocol/scenarios/action/components/actionForm/testplan-run.go
+++ b/modules/openapi/component-protocol/scenarios/action/components/actionForm/testplan-run.go
@@ -47,6 +47,8 @@ func testPlanRun(ctx context.Context, c *apistructs.Component, scenario apistruc
 	// get testplan
 	testPlanRequest := apistructs.TestPlanV2PagingRequest{
 		ProjectID: uint64(projectId),
+		PageNo:    1,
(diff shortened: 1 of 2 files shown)


modules/openapi/component-protocol/scenarios/action/components/actionForm/testplan-run.go
```diff
@@ -47,6 +47,8 @@ func testPlanRun(ctx context.Context, c *apistructs.Component, scenario apistruc
 	// get testplan
 	testPlanRequest := apistructs.TestPlanV2PagingRequest{
 		ProjectID: uint64(projectId),
+		PageNo:    1,
+		PageSize:  999,
```
sfwn, line 51: Does backend allow big pageSize?

sfwn (changes requested):

modules/openapi/component-protocol/scenarios/action/components/actionForm/testplan-run.go
kakj-go: yes

sfwn (approved):

PR #1998, opened 2021-09-18
fix test sum

recallsong:
#### What type of this PR

/kind bug

fix test sum

#### Specified Reviewers:

/assign @hoozecn


hoozecn (approved):

hoozecn (changes requested):
test failed

PR #2164, opened 2021-09-28
Update project check permission with org id where the project is located

shuofan:
#### What type of this PR
bug

- Fixes #your-issue_number
- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=218949&issueFilter__urlQuery=eyJzdGF0ZUJlbG9uZ3MiOlsiT1BFTiIsIldPUktJTkciLCJXT05URklYIiwiUkVPUEVOIiwiUkVTT0xWRUQiXX0%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&issueViewGroup__urlQuery=eyJ2YWx1ZSI6InRhYmxlIiwiY2hpbGRyZW5WYWx1ZSI6eyJrYW5iYW4iOiJkZWFkbGluZSJ9fQ%3D%3D&iterationID=0&type=BUG)

#### Specified Reviewers:

/assign @sfwn


modules/core-services/endpoints/project.go
```diff
@@ -81,7 +81,7 @@ func (e *Endpoints) UpdateProject(ctx context.Context, r *http.Request, vars map
 		return apierrors.ErrUpdateProject.NotLogin().ToResp(), nil
 	}
 	orgIDStr := r.Header.Get(httputil.OrgHeader)
-	orgID, err := strutil.Atoi64(orgIDStr)
+	_, err = strutil.Atoi64(orgIDStr)
```
Effet, line 84: We should compare the equality of orgId in header with orgId in projectInfo. If they are not equal, a appropriate error should be returned (it is also permission denied because user beyond the boundary of org).

Effet (changes requested):

modules/core-services/endpoints/project.go
shuofan: make sense

Effet (approved):

PR #2186, opened 2021-09-28
Fix the problem of inaccurate execution time of automatic test

kakj-go:
#### What type of this PR
/kind bug

Fix the problem of inaccurate execution time of automatic test

### DIFF
--- a/modules/pipeline/pipengine/actionexecutor/plugins/apitest/apitest.go
+++ b/modules/pipeline/pipengine/actionexecutor/plugins/apitest/apitest.go
@@ -57,7 +57,40 @@ func (d *define) Create(ctx context.Context, task *spec.PipelineTask) (interface
 }
 
 func (d *define) Start(ctx context.Context, task *spec.PipelineTask) (interface{}, error) {
-	logic.Do(ctx, task)
+	executorDoneCh := ctx.Value(spec.MakeTaskExecutorCtxKey(task)).(chan interface{})
+	if executorDoneCh == nil {
+		return nil, fmt.Errorf("wait: failed to get exector channel, pipelineID: %d, taskID: %d", task.PipelineID, task.ID)
+	}
+	go func() {
+
+		var status = apistructs.PipelineStatusFailed
+		defer func() {
+			if r := recover(); r != nil {
+				fmt.Printf("api-test logic do panic recover:%s", r)
+				executorDoneCh <- apistructs.PipelineStatusDesc{Status: status}
+			}
+		}()
+
+		logic.Do(ctx, task)
+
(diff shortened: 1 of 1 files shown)


modules/pipeline/pipengine/actionexecutor/plugins/apitest/apitest.go
```diff
@@ -57,7 +57,32 @@ func (d *define) Create(ctx context.Context, task *spec.PipelineTask) (interface
...
+	executorDoneCh := ctx.Value(spec.MakeTaskExecutorCtxKey(task)).(chan interface{})
+	if executorDoneCh == nil {
+		return nil, fmt.Errorf("wait: failed to get exector channel, pipelineID: %d, taskID: %d", task.PipelineID, task.ID)
+	}
+	go func() {
+		logic.Do(ctx, task)
+
+		latestTask, err := d.dbClient.GetPipelineTask(task.ID)
+		if err != nil {
+			fmt.Printf("failed to query latest task, err: %v \n", err)
```
sfwn, line 69: if get db error, this api-test task will never stop?

modules/pipeline/pipengine/actionexecutor/plugins/apitest/apitest.go
```diff
@@ -57,7 +57,32 @@ func (d *define) Create(ctx context.Context, task *spec.PipelineTask) (interface
 }
 
 func (d *define) Start(ctx context.Context, task *spec.PipelineTask) (interface{}, error) {
-	logic.Do(ctx, task)
+	executorDoneCh := ctx.Value(spec.MakeTaskExecutorCtxKey(task)).(chan interface{})
+	if executorDoneCh == nil {
+		return nil, fmt.Errorf("wait: failed to get exector channel, pipelineID: %d, taskID: %d", task.PipelineID, task.ID)
+	}
+	go func() {
+		logic.Do(ctx, task)
```
sfwn, line 65: What will happen if panic at logic.Do?

sfwn (changes requested):

sfwn:
Add your test screenshot here pls.

modules/pipeline/pipengine/actionexecutor/plugins/apitest/apitest.go
kakj-go: The query error will not be running all the time. The latest status query here is just to terminate the operation early. This code can be removed because it will always be queried in the status method of the executer, but the scheduled time of the status is incremented. Reach 10s, if the asynchronous execution does not return early, the execution time will be longer

kakj-go:
> Add your test screenshot here pls.

image

image

pipeline_tasks table cost_time_sec has value

sfwn (approved):

PR #3661, opened 2022-01-10
fix: manual-test recognize member 'unassigned' as id

sfwn:

fix: manual-test recognize member 'unassigned' as id

- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=269348&issueFilter__urlQuery=eyJ0aXRsZSI6IuacquaMh%2BWumiIsInN0YXRlcyI6WzQ0MTIsNDUzOCw0NDEzLDQ0MTQsNDQxNSw0NDE2XX0%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&iterationID=772&type=BUG)

#### Specified Reviewers:

/assign @Effet 

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |  Fixed an bug that manual-test could not filter out "unspecified" users          |
| 🇨🇳 中文    |   修复了手动测试无法过滤出"未指定"用户的问题           |

### DIFF
--- a/modules/dop/dao/testplan_testcase_relation.go
+++ b/modules/dop/dao/testplan_testcase_relation.go
@@ -25,6 +25,7 @@ import (
 	"github.com/erda-project/erda/modules/dop/services/apierrors"
 	"github.com/erda-project/erda/pkg/database/dbengine"
 	"github.com/erda-project/erda/pkg/strutil"
+	"github.com/erda-project/erda/pkg/ucauth"
 )
 
 // TestPlanCaseRel
@@ -495,5 +496,17 @@ func setDefaultForTestPlanCaseRelPagingRequest(req *apistructs.TestPlanCaseRelPa
 	}
 	if req.PageSize == 0 {
 		req.PageSize = 20
+	}
+
+	// set unassigned ids
+	for i, executorID := range req.ExecutorIDs {
+		if ucauth.USERID(executorID).IsUnassigned() {
+			req.ExecutorIDs[i] = ""
+		}
+	}
+	for i, updaterID := range req.UpdaterIDs {
+		if ucauth.USERID(updaterID).IsUnassigned() {
+			req.UpdaterIDs[i] = ""
+		}
 	}
 }
--- a/modules/dop/dao/testplan_testcase_relation.go
+++ b/modules/dop/dao/testplan_testcase_relation.go
@@ -324,7 +324,7 @@ func (client *DBClient) PagingTestPlanCaseRelations(req apistructs.TestPlanCaseR
 	}
 	// updater
 	if len(req.UpdaterIDs) > 0 {
-		baseSQL = baseSQL.Where("`tc`.`updater_id` IN (?)", req.UpdaterIDs)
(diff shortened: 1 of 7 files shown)


pkg/ucauth/unassigned.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func (u USERID) IsUnassigned() bool {
+	return strings.EqualFold(u.String(), UnassignedUserID.String())
+}
+
+func PolishUnassignedAsEmptyStr(userIDs []string) (result []string) {
+	for _, userID := range userIDs {
+		polishedUserID := userID
+		if strings.EqualFold(userID, UnassignedUserID.String()) {
```
Effet, line 33: why you not use `IsUnassigned`?

Effet (changes requested):

pkg/ucauth/unassigned.go
sfwn: Forgot this single method when add batch handler. 
Adjusted.

modules/dop/endpoints/testcase.go
```diff
@@ -212,32 +212,6 @@ func (e *Endpoints) PagingTestCases(ctx context.Context, r *http.Request, vars m
...
-				mapOfupdaterIDs[member.UserID] = true
-			}
-		}
-		for _, value := range mapOfupdaterIDs {
-			if !value {
-				return nil, apierrors.ErrPagingTestCases.AccessDenied()
-			}
-		}
-	}
-
```
Effet, line 240: why you remove this?

modules/dop/dao/testplan_testcase_relation.go
```diff
@@ -323,7 +324,7 @@ func (client *DBClient) PagingTestPlanCaseRelations(req apistructs.TestPlanCaseR
 	}
 	// updater
 	if len(req.UpdaterIDs) > 0 {
-		baseSQL = baseSQL.Where("`tc`.`updater_id` IN (?)", req.UpdaterIDs)
+		baseSQL = baseSQL.Where("`rel`.`updater_id` IN (?)", req.UpdaterIDs)
```
Effet, line 327: why change tc to rel

Effet (changes requested):

modules/dop/endpoints/testcase.go
sfwn: Direct reason: if updaterIDs contains `unassigned`, you will occur an error of 403.
In fact, no need check updater's permission in paging request.

modules/dop/dao/testplan_testcase_relation.go
sfwn: Both `TestPlanCaseRelation` and `TestCase` have `updater_id` field.
In method `PagingTestPlanCaseRelations`, we should use `rel`'s `updater_id`.

Effet (approved):

PR #3666, opened 2022-01-11
fix: batch runtime scale API return data may be nil

wangzhuzhen:
 bugfix: batch runtime scale API return data may be nil

#### Specified Reviewers:

/assign @sixther-dc @iutx 

#### ChangeLog
bugfix: batch runtime scale API return data may be nil

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |     bugfix: batch runtime scale API return data may be nil         |
| 🇨🇳 中文    |       问题修复：批量 runtimes scale 处理 API 可能返回空值导致空指针错误        |


sixther-dc (approved):
/approve

PR #3668, opened 2022-01-11
fix: transfer data struct

wangzhuzhen:
batch runtimes scale return data is map[string]interface{}, need transfer to specifical data struct  

#### Specified Reviewers:

/assign @sixther-dc @iutx 

#### ChangeLog
Bugfix： Fix the  data struct assert bug for  map[string]interface{}

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |      Fix the  data struct assert bug for  map[string]interface{}         |
| 🇨🇳 中文    |         修复数据结构断言 panic 问题     |

### DIFF
--- a/bundle/orchestrator.go
+++ b/bundle/orchestrator.go
@@ -15,6 +15,7 @@
 package bundle
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


sixther-dc (approved):
/approve

PR #3797, opened 2022-01-18
fix: alert and dashboard audit

Counterflowwind:
alert and dashboard audit

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |       alert and dashboard audit       |
| 🇨🇳 中文    |        修改审计日志      |


liuhaoyang (approved):

pkg/common/apis/header.go
```diff
@@ -77,6 +77,10 @@ func GetUserID(ctx context.Context) string {
 	return GetHeader(ctx, "user-id")
 }
 
+func GetMSPInfo(ctx context.Context, key string) string {
+	return GetHeader(ctx, key)
```
sfwn, line 81: you should use a more common method name, such as `GetValueByKey` or `GetValue`.

liuhaoyang: You are right, I will create a new pr to fix this method naming.

PR #3960, opened 2022-01-26
fix: cancel project pipeline error

kakj-go:
Solve some abnormal situations, such as when the execution state of the pipeline is not synchronized with the state defined by the pipeline, the user will report an error when canceling the pipeline

- [Erda Cloud Issue Link](https://terminus-org.app.terminus.io/erda/dop/projects/387/issues/gantt?filter__urlQuery=eyJpdGVyYXRpb24iOls4ODMsNzcyLDg4MV0sIm1lbWJlciI6WyIxMDAwNTYwIl19&id=279352&iterationID=772&pId=0&type=BUG)

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |       Solve some abnormal situations, such as when the execution state of the pipeline is not synchronized with the state defined by the pipeline, the user will report an error when canceling the pipeline       |
| 🇨🇳 中文    |        解决一些异常情况，例如当流水线的执行状态与流水线定义的状态不同步时，用户在取消流水线时会返回错误     |


modules/pipeline/providers/definition/pipeline_definition.go
```diff
@@ -235,6 +235,7 @@ func PipelineDefinitionToPb(pipelineDefinition *db.PipelineDefinition) *pb.Pipel
 		TimeUpdated:      timestamppb.New(pipelineDefinition.TimeUpdated),
 		StartedAt:        timestamppb.New(pipelineDefinition.StartedAt),
 		EndedAt:          timestamppb.New(pipelineDefinition.EndedAt),
+		PipelineId:       int64(pipelineDefinition.PipelineID),
```
sfwn, line 238: bad Id. use ID in proto and set json tag please.

modules/dop/providers/projectpipeline/service_impl.go
```diff
@@ -720,22 +720,29 @@ func (p *ProjectPipelineService) Cancel(ctx context.Context, params deftype.Proj
...
+	if pipelineInfo.Status.IsRunningStatus() {
+		var req apistructs.PipelineCancelRequest
+		req.PipelineID = uint64(definition.PipelineId)
+		req.IdentityInfo = params.IdentityInfo
+		err = p.bundle.CancelPipeline(req)
+		if err != nil {
+			return nil, apierrors.ErrCancelProjectPipeline.InternalError(err)
+		}
+
+		_, err = p.PipelineDefinition.Update(context.Background(), &dpb.PipelineDefinitionUpdateRequest{PipelineDefinitionID: definition.ID, Status: string(apistructs.PipelineStatusStopByUser), PipelineId: definition.PipelineId})
```
sfwn, line 737: maybe you can do this inside cancelPipeline?

sfwn (changes requested):

modules/dop/providers/projectpipeline/service_impl.go
kakj-go: Interfaces such as execution, execution at repeated failures, Retry, cancel execution, and start timing should be considered in the pipeline component, which should be reconstructed with another pr

api/proto/core/pipeline/definition/definition.proto
```diff
@@ -65,7 +65,7 @@ message PipelineDefinition {
   string fileName = 18;
   string pipelineSourceId = 19;
```
sfwn, line 66: and this one

sfwn (changes requested):

api/proto/core/pipeline/definition/definition.proto
kakj-go: One PR only does one thing, and the next PR completes the standard naming

sfwn (approved):

PR #4072, opened 2022-02-16
fix: fix compatibility for old service volumes

wangzhuzhen:
Compatibility for old service volumes and new service volume defines

#### Specified Reviewers:

/assign @sixther-dc 

#### ChangeLog
Bugfix： Fix the bug that can not compatible with old volume defines

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |      Compatibility for old service volumes and new service volume defines        |
| 🇨🇳 中文    |      Compatibility for old service volumes and new service volume defines        |


sixther-dc (approved):
/approve

PR #4176, opened 2022-02-28
fix: fix get containerID out of range fatal

wangzhuzhen:
Get containerID from Pod Status may get unknown container which containerID may be nil. 

#### Specified Reviewers:

/assign @sixther-dc 

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |         Fix the bug that get containerID from Pod Status may get unknown container which containerID is nil.,which result in out of range fatal     |
| 🇨🇳 中文    |      修复了 Pod Status 中包含未知状态的容器导致容器ID为空触发数组越界        |

/cherry-pick 1.6-alpha.4


sixther-dc (approved):
/approve

PR #4273, opened 2022-03-04
fix(dop): When the pipeline is GC, if the definition is bound to the pipeline_ ID does not perform GC operations

kakj-go:
Solve the problem that the pipeline bound by the definition is GC, which causes the user to view the execution details and report an error

- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=289045&issueFilter__urlQuery=eyJzdGF0ZXMiOls0NDEyLDQ1MzgsNDQxMyw0NDE0LDQ0MTUsNDQxNl0sImFzc2lnbmVlSURzIjpbIjEwMDA1NjAiXX0%3D&issueTable__urlQuery=eyJwYWdlTm8iOjEsICJwYWdlU2l6ZSI6MTB9&iterationID=1115&type=BUG)

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |       Solve the problem that the pipeline bound by the definition is GC, which causes the user to view the execution details and report an error       |
| 🇨🇳 中文    |       解决由于 definition 绑定的 pipeline 被 gc，导致用户查看执行明细报错       |

test image:

change bind pipeline_id time info:
image

restart pipeline gc info:
image

### DIFF
--- a/modules/pipeline/pipengine/reconciler/dbgc.go
+++ b/modules/pipeline/pipengine/reconciler/dbgc.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strconv"
+	"strings"
 	"time"
 
 	v3 "github.com/coreos/etcd/clientv3"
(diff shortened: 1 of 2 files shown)


modules/pipeline/pipengine/reconciler/dbgc.go
```diff
@@ -239,6 +240,23 @@ func (r *Reconciler) DoDBGC(pipelineID uint64, gcOption apistructs.PipelineGCDBO
...
 	}
 
+	// If the pipeline is referenced by the definition and the record is GC
+	// the definition cannot see the execution record.
+	// There are two ways:
+	// 		one is to update the definition information contact binding
+	//	 	and the other is not to go to the GC pipeline record
+	// It is decided not to record this pipeline
+	if strings.TrimSpace(p.PipelineDefinitionID) != "" {
+		definition, err := r.dbClient.GetPipelineDefinitionByPipelineID(p.PipelineID)
```
littlejiancc, line 250: It is better to return the ‘has’ parameter or distinguish whether it exists or not.
Otherwise, a lot of non-existent logs will be printed.

kakj-go: done

sfwn (approved):

PR #4572, opened 2022-04-08
fix(pipeline): When creating a definition, a cron info is created asynchronously

kakj-go:
When creating a definition, a scheduled task is created asynchronously

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |      When creating a definition, a scheduled task is created asynchronously        |
| 🇨🇳 中文    |       创建定义的时候异步创建定时任务       |

### DIFF
--- a/modules/dop/providers/projectpipeline/service_impl.go
+++ b/modules/dop/providers/projectpipeline/service_impl.go
@@ -234,8 +234,8 @@ func (p *ProjectPipelineService) Create(ctx context.Context, params *pb.CreatePr
 		return nil, apierrors.ErrCreateProjectPipeline.InternalError(fmt.Errorf(sourceCheckResult.Message))
 	}
 
-	p.pipelineSourceType = NewProjectSourceType(params.SourceType)
-	sourceReq, err := p.pipelineSourceType.GenerateReq(ctx, p, params)
+	pipelineSourceType := NewProjectSourceType(params.SourceType)
+	sourceReq, err := pipelineSourceType.GenerateReq(ctx, p, params)
 	if err != nil {
 		return nil, apierrors.ErrCreateProjectPipeline.InternalError(err)
 	}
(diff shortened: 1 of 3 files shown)


modules/dop/providers/projectpipeline/service_impl.go
```diff
@@ -257,14 +257,22 @@ func (p *ProjectPipelineService) Create(ctx context.Context, params *pb.CreatePr
...
 		},
 		Ref: sourceRsp.PipelineSource.Ref,
 	})
 	if err != nil {
 		return nil, apierrors.ErrCreateProjectPipeline.InternalError(err)
 	}
 
+	// When creating a definition, a scheduled task is created asynchronously
+	go func() {
+		err := p.createCronIfNotExist(definitionRsp.PipelineDefinition, pipelineSourceType)
```
sfwn, line 270: any compensate mechanism if create failed?

sfwn (changes requested):

modules/dop/providers/projectpipeline/service_impl.go
kakj-go: There is no open transaction, even if it is not asynchronous, there is an inconsistency problem

sfwn: what will happen if cron create failed? The invoker **DO NOT** know whether the cron success or not, and you return success even when cron failed.
It will cause cron task cannot be executed as expected, and it's serious enough.

modules/dop/providers/projectpipeline/service_impl.go
```diff
@@ -286,6 +294,39 @@ func (p *ProjectPipelineService) Create(ctx context.Context, params *pb.CreatePr
...
 
+func (p *ProjectPipelineService) createCronIfNotExist(definition *dpb.PipelineDefinition, projectPipelineType ProjectSourceType) error {
+	extraString := projectPipelineType.GetPipelineCreateRequestV2()
+	var extra apistructs.PipelineDefinitionExtraValue
+	err := json.Unmarshal([]byte(extraString), &extra)
+	if err != nil {
+		return err
+	}
+
+	crons, err := p.PipelineCron.CronPaging(context.Background(), &cronpb.CronPagingRequest{
```
sfwn, line 305: I think you should also search existed-cron by definition id.

sfwn (commented):
add your erda issue link if you have.

modules/dop/providers/projectpipeline/service_impl.go
kakj-go: This method is called when the definition is created, there can not be have cron that contains the definition just created

modules/dop/providers/projectpipeline/service_impl.go
kakj-go: I changed to synchronous call

sfwn (approved):

PR #4622, opened 2022-04-15
fix: return null if there is no release with target tags

CraigMChen:

fix: return null if there is no release with target tags

- Fixes #your-issue_number

#### Specified Reviewers:

/assign @sixther-dc 

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English | fix: return null if there is no release with target tags |
| 🇨🇳 中文    | 若没有制品符合筛选标签，则返回空 |


sixther-dc (approved):
/approve

PR #4971, opened 2022-05-30
fix: use changed-files action to handle push event

sfwn:

We use pr number from workflow context before, so it fails when squash merged (it's a push event, no pr id).

image

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |   use changed-files action to handle push event           |
| 🇨🇳 中文    |     使用 changed-files Action 来处理 Push 事件         |


sfwn:
It works.
image

PR #5208, opened 2022-07-08
fix(orchestrator): process hpa event may get nil from orgCache

wangzhuzhen:
for hpa events in edge cluster, center cluster orchestrator orgCache may not cached the Org info for that hpa,  so orgCache may get nil

#### Specified Reviewers:

/assign @sixther-dc 

#### ChangeLog

Bugfix： Fix the bug that the orgCache may not cached the Org  info, need set default org Locale    （修复了处理 HPA 事件时获取Org 信息失败导致获取 Org 的 Locate 为空引发空指针的问题）

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |    Fix the bug that the orgCache may not cached the Org  info, so need set default org Locale  |
| 🇨🇳 中文    |      修复了处理 HPA 事件时获取Org 信息失败导致获取 Org 的 Locate 为空引发空指针的问题        |

2.3-alpha

### DIFF
--- a/internal/tools/orchestrator/scheduler/executor/plugins/k8s/instanceinfosync/sync.go
+++ b/internal/tools/orchestrator/scheduler/executor/plugins/k8s/instanceinfosync/sync.go
@@ -363,11 +363,14 @@ func (s *Syncer) watchSyncHPAEvent(ctx context.Context) {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


sixther-dc (approved):
/approve

PR #5401, opened 2022-08-12
fix: get_services_pods for running pods exclude terminated Pod

wangzhuzhen:

get_services_pods for running pods need exclude terminated pods. For now, if pod evicted, there may be many Failed pods which displayed.

#### Specified Reviewers:

/assign @luobily @sixther-dc @iutx 

#### ChangeLog
Bugfix： Fix the bug that get_services_pods for running pods need exclude terminated pods （修复了get_services_pods 接口没有排除终止或者未知状态的 Pods ）

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |     Fix the bug that get_services_pods for running pods need exclude terminated pods         |
| 🇨🇳 中文    |        修复了get_services_pods 接口没有排除终止或者未知状态的 Pods      |

### DIFF
--- a/internal/tools/orchestrator/endpoints/instance.go
+++ b/internal/tools/orchestrator/endpoints/instance.go
@@ -42,6 +42,11 @@ const (
 	INSTANCE_REQUEST_TYPE_APPLICATION InstanceRequestType = "application"
 	INSTANCE_REQUEST_TYPE_SERVICE     InstanceRequestType = "service"
 	INSTANCE_REQUEST_TYPE_RUNTIME     InstanceRequestType = "runtime"
+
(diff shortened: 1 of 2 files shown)


apistructs/container.go
```diff
@@ -150,6 +150,7 @@ type Pod struct {
 	IPAddress     string         `json:"ipAddress"`
 	Host          string         `json:"host"`
 	Phase         string         `json:"phase"`
+	IsPodHealthy  bool           `json:"isPodHealthy"`
```
luobily, line 153: Phase and IsPodHealthy can be combined, do not directly reveal the phase of K8S.

sixther-dc (approved):
/approve

PR #5439, opened 2022-08-18
fix: get service_pods return different runtime's pods 

wangzhuzhen:

get service_pods return different runtime's pods in the same k8s namespace but with different runtimeID

#### Specified Reviewers:

/assign @sixther-dc @iutx 

#### ChangeLog

Bugfix： Fix the bug that get service_pods return different runtime's pods in the same k8s namespace but with different runtimeID（修复了 get service_pods 接口返回同一个namespace 下具有相同 serviceName 的不同 runtimeID 的 pods ）

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |      Fix the bug that get service_pods return different runtime's pods in the same k8s namespace but with different runtimeID        |
| 🇨🇳 中文    |      修复了 get service_pods 接口返回同一个namespace 下具有相同 serviceName 的不同 runtimeID 的 pods         |


sixther-dc (approved):
/approve

PR #6120, opened 2023-11-02
feat(ai-proxy): bailian qwenv1 support both stream&non-stream request

sfwn:

bailian qwenv1 support both stream&non-stream request

- [Erda Cloud Issue Link](https://erda.cloud/terminus/dop/projects/1265/issues/all?id=532809&iterationID=13370&type=TASK)

#### Specified Reviewers:

/assign @wangzhuzhen 

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |   bailian qwenv1 support both stream&non-stream request           |
| 🇨🇳 中文    |  阿里云百炼通义千问模型支持流式和非流式请求            |

### DIFF
--- a/internal/apps/ai-proxy/common/ctxhelper/stream.go
+++ b/internal/apps/ai-proxy/common/ctxhelper/stream.go
@@ -12,24 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package vars
+package ctxhelper
 
 import (
-	"strings"
+	"context"
+	"sync"
+
+	"github.com/erda-project/erda/internal/apps/ai-proxy/vars"
+	"github.com/erda-project/erda/pkg/reverseproxy"
 )
 
-func ConcatBearer(v string) string {
-	return "Bearer " + v
+func GetIsStream(ctx context.Context) (bool, bool) {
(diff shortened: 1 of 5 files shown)


internal/apps/ai-proxy/common/ctxhelper/stream.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/erda-project/erda/internal/apps/ai-proxy/vars"
+	"github.com/erda-project/erda/pkg/reverseproxy"
+)
+
+func GetIsStream(ctx context.Context) (bool, bool) {
```
wangzhuzhen, line 25: why return 2 bool？ only use 1 bool in context.

sfwn: Yes, the first bool means the value of `isStream`, and the second bool is used to confirm whether it exists or not in context.
Other Get/Put method under `ctxhelper` pkg follow the same rule.

wangzhuzhen: OK

PR #6360, opened 2024-06-06
fix(gh-ci):  set safe.directory at Dockerfile side to prevent "dubious ownership" errors

sfwn:
#### What this PR does / why we need it:

set safe.directory at Dockerfile side to prevent "dubious ownership" errors

#### Which issue(s) this PR fixes:

- Fixes #your-issue_number
- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=604917&iterationID=12783&tab=ALL&type=TASK)

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |    fix github ci: set safe.directory at Dockerfile side to prevent "dubious ownership" errors          |
| 🇨🇳 中文    |   修复 github ci: 在 Dockerfile 侧设置 safe.directory 来避免 "dubious ownership" 错误          |

PR #6361, opened 2024-06-06
fix(gh-ci): use `--system` instead of `--global` to make it work for all users

sfwn:

use `--system` instead of `--global` to make it work for all users

- Fixes #your-issue_number
- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=604917&iterationID=12783&tab=ALL&type=TASK)

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English | use `--system` instead of `--global` to make it work for all users to prevent "dubious ownership" error          |
| 🇨🇳 中文    |  使用 `--system` 而不是 `--global` 使所有用户生效，来避免 "dubious ownership" 错误           |


sfwn:
previous pr: #6360

PR #6522, opened 2025-05-22
feat(ai-proxy): support set extra-body for some models

sfwn:

ai-proxy support set extra-body for some models.

such as: qwen3-235b-a22b non-stream set unofficial `enable-thinking=false` as default.

default configs configured at model's public metadata:

image

- [Erda Cloud Issue Link](https://erda.cloud/erda/dop/projects/387/issues/all?id=759363&iterationID=12783&type=TASK)

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |   ai-proxy  support set extra-body for some models          |
| 🇨🇳 中文    |    ai-proxy 支持为模型设置额外 request body          |

### DIFF
--- a/internal/apps/ai-proxy/filters/extra-body/json_body.go
+++ b/internal/apps/ai-proxy/filters/extra-body/json_body.go
@@ -1,0 +1,67 @@
+// Copyright (c) 2021 Terminus, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


internal/apps/ai-proxy/filters/extra-body/json_body.go
```diff
@@ -0,0 +1,67 @@
...
+			body[k] = v
+		}
+	}
+	return nil
+}
+
+// unmarshalJSONFromString unmarshal json string to map[string]any
+func unmarshalJSONFromString(s string) (map[string]any, error) {
+	if s == "" {
+		return nil, nil
```
iutx, line 60: It would be more appropriate to return an empty initialized map[string]any, although ranging over a nil map does not cause a panic.

iutx (commented):
@sfwn

internal/apps/ai-proxy/filters/extra-body/json_body.go
sfwn: yes, it's a feature in [go spec](https://go.dev/ref/spec).

`A nil map is equivalent to an empty map except that no elements may be added.`

iutx: Adding an element will cause panic unless the map is read-only in the future.

PR #6773, opened 2026-04-22
feat(cli): add release cli tooling

iutx:
add release cli tooling

- Fixes #your-issue_number

#### Specified Reviewers:

/assign @luobily 

#### ChangeLog

| Language | Changelog |
| --------- | ------------ |
| 🇺🇸 English |    add release cli tooling          |
| 🇨🇳 中文    |     增加 release-cli 能力         |

Add comment like `/cherry-pick release/1.0` when this PR is merged.

> For details on the cherry pick process, see the [cherry pick requests](https://github.com/erda-project/erda/blob/master/CONTRIBUTING.md#how-to-cherry-pick-a-merged-pr) section under [CONTRIBUTING.md](https://github.com/erda-project/erda/blob/master/CONTRIBUTING.md).


luobily (changes requested):
Thanks for adding the release CLI tooling. I found a couple of release-path issues that should be fixed before merging.

## Findings

1. **High: `release-cli` Makefile targets currently fail to compile**

`Makefile` invokes the tool as:

```make
go run tools/release-cli/main.go publish --version "${VERSION}" --dir "${PROJ_PATH}/bin"
go run tools/release-cli/main.go prune --apply
```

But `main.go` depends on symbols from other files in the same package, such as `newRootCmd` in `root.go`. Running a single file does not include those files, so the target fails with:

```text
# command-line-arguments
tools/release-cli/main.go:31:9: undefined: newRootCmd
```

Please switch these to package-mode execution:

```make
go run ./tools/release-cli publish --version "${VERSION}" --dir "${PROJ_PATH}/bin"
go run ./tools/release-cli prune --apply
```

I verified that `go run ./tools/release-cli --help` works, while `go run tools/release-cli/main.go --help` fails.

2. **High: New OSS artifact layout is incompatible with the existing installer**

The new release code publishes archives under paths like:

```text
cli/{goos}/{goarch}/erda-cli-{version}.tar.gz
```

But `tools/cli/install.sh` still downloads:

```text
https://erda-release.oss-cn-hangzhou.aliyuncs.com/cli/$OS/erda-$Version
```

Those layouts differ in OS naming (`mac` vs `darwin`), architecture path, file name, and archive format. Once this release path is used, existing installer users will get a 404 or a non-executable archive written to `/usr/local/bin/erda-cli`.

Please either update `install.sh` to resolve the new channel/version manifest, select OS/arch, download and extract the archive, or keep publishing backward-compatible bare binaries at the old paths.

3. **Medium: Published manifest URLs ignore configurable OSS endpoint/bucket**

`buildManifest` always uses `release.DefaultBaseURL` when setting `Manifest.URL`, even though the CLI supports `OSS_ENDPOINT` and `OSS_BUCKET_NAME` overrides. If this tool is used with a staging bucket or custom endpoint, the uploaded manifest will still point clients to the production `erda-release` bucket.

Please derive the public base URL from the actual publish configuration, or add an explicit `OSS_BASE_URL`/`--base-url` option and use that in the manifest.

4. **Medium: `prune` can leave channel manifests pointing at deleted artifacts**

`pruneReleases` deletes old version manifests/artifacts and rebuilds `*-versions.json`, but it does not update the channel manifest (`alpha.json`, `beta.json`, `stable.json`). If the channel manifest currently points at a version that gets pruned, clients using the channel manifest can continue receiving a URL for an artifact that was just deleted.

Please update the channel manifest to the first retained version during prune, or explicitly protect the current channel manifest target from deletion.

5. **Low: Publish and prune use different platform matrices**

Publishing currently targets only `darwin/arm64` and `linux/amd64`, while prune scans `darwin/linux/windows` across both `amd64/arm64`. That makes prune manage object prefixes that this publish flow does not create. It is safer to share one supported target matrix, or require extra platforms to be passed explicitly.

## Verification

I tested the focused packages on the PR snapshot:

```text
go test ./tools/cli/release ./tools/release-cli
ok
```

The package tests pass, but the Makefile invocation issue above is directly reproducible with `go run tools/release-cli/main.go --help`.

Makefile
```diff
@@ -152,23 +166,30 @@ prepare-cli:
...
-	go run tools/upload-cli/main.go ${ACCESS_KEY_ID} ${ACCESS_KEY_SECRET} cli/mac/erda "${PROJ_PATH}/bin/erda-cli"
-	go run tools/upload-cli/main.go ${ACCESS_KEY_ID} ${ACCESS_KEY_SECRET} cli/linux/erda "${PROJ_PATH}/bin/erda-cli-linux"
+.PHONY: release-cli-build
+release-cli-build: prepare-cli
+	$(call build_cli_binary,darwin,arm64,${PROJ_PATH}/bin/erda-cli)
+	$(call build_cli_binary,linux,amd64,${PROJ_PATH}/bin/erda-cli-linux)
+
+.PHONY: release-cli-publish
+release-cli-publish: release-cli-build
+	@go run tools/release-cli/main.go publish --version "${VERSION}" --dir "${PROJ_PATH}/bin"
```
luobily: This target currently fails before it can publish. `go run tools/release-cli/main.go ...` compiles only `main.go`, but `main.go` calls `newRootCmd` from `root.go`, so the command fails with `undefined: newRootCmd`. Please run the package instead, e.g. `go run ./tools/release-cli publish --version "${VERSION}" --dir "${PROJ_PATH}/bin"`. The same applies to the prune target below.

Makefile
```diff
@@ -152,23 +166,30 @@ prepare-cli:
...
+
+.PHONY: release-cli
+release-cli: release-cli-publish
+
+.PHONY: release-cli-prune
+release-cli-prune:
+	@go run tools/release-cli/main.go prune --apply
```
luobily: Same single-file `go run` issue as the publish target: this should be `go run ./tools/release-cli prune --apply`, otherwise `release-cli-prune` fails to compile because the other files in `tools/release-cli` are not included.

tools/release-cli/publish.go
```diff
@@ -0,0 +1,243 @@
...
+	if _, err := io.Copy(sum, file); err != nil {
+		return nil, err
+	}
+
+	return &release.Manifest{
+		Version:     target.version,
+		Channel:     target.channel,
+		OS:          target.goos,
+		Arch:        target.goarch,
+		URL:         release.DefaultBaseURL + "/" + release.ArtifactObjectName(target.goos, target.goarch, target.version),
```
luobily: This manifest URL ignores the configured `OSS_ENDPOINT` and `OSS_BUCKET_NAME`. If the tool publishes to a staging bucket or custom endpoint, the manifest still points clients at the production `erda-release` bucket. Please derive the public base URL from publish options, or add an explicit `--base-url` / `OSS_BASE_URL` and use that here.

tools/release-cli/prune.go
```diff
@@ -0,0 +1,254 @@
...
+				return err
+			}
+		}
+		index := &release.VersionIndex{
+			Channel:  plan.channel,
+			OS:       plan.goos,
+			Arch:     plan.goarch,
+			Versions: plan.retained,
+		}
+		indexObjectName := release.ChannelVersionsObjectName(plan.goos, plan.goarch, plan.channel)
```
luobily, line 157: Prune rebuilds `*-versions.json`, but it never updates the channel manifest (`alpha.json`, `beta.json`, `stable.json`). If the channel manifest points at a version that gets deleted, clients using the channel manifest can continue receiving a URL for a removed artifact. Please update the channel manifest to `plan.retained[0]` during prune, or protect the current channel manifest target from deletion.

tools/release-cli/config.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+const (
+	defaultOSSEndpoint   = "oss-cn-hangzhou.aliyuncs.com"
+	defaultOSSBucketName = "erda-release"
+	artifactPrefix       = "cli"
+)
+
+var (
+	defaultOSList   = []string{"darwin", "linux", "windows"}
```
luobily, line 30: Publish and prune use different platform matrices: publish only creates `darwin/arm64` and `linux/amd64`, while prune scans `darwin`, `linux`, and `windows` across both `amd64` and `arm64`. That makes prune manage prefixes this release flow does not create. Please share one supported target matrix or make extra prune platforms explicit.

luobily (changes requested):
Thanks for adding the release CLI tooling. I split the review into line-level comments where the issues are directly on changed lines. One additional release compatibility issue is not attached inline because `tools/cli/install.sh` is not part of this PR diff: the new publisher writes archives under `cli/{goos}/{goarch}/erda-cli-{version}.tar.gz`, while the existing installer still downloads `cli/$OS/erda-$Version` with `mac` naming and writes the response directly to `/usr/local/bin/erda-cli`. Please either update the installer to use the new manifests and extract archives, or keep publishing backward-compatible bare binaries at the old paths.

Verification performed on the PR snapshot:

```text
go test ./tools/cli/release ./tools/release-cli
ok
```

The focused tests pass, but the Makefile invocation issue is directly reproducible with `go run tools/release-cli/main.go --help`.

tools/release-cli/config.go
iutx: This is intentional. We expect to expand releaseTargets in future releases, and prune is deliberately pre-provisioned to handle the broader platform matrix ahead of time.

tools/release-cli/publish.go
```diff
@@ -0,0 +1,261 @@
...
+		return err
+	}
+	defer cleanup()
+
+	manifest, err := buildManifest(target, archivePath)
+	if err != nil {
+		return err
+	}
+
+	artifactObject := release.ArtifactObjectName(target.goos, target.goarch, target.version)
```
luobily, line 180: This switches the published artifact to `cli/{goos}/{goarch}/erda-cli-{version}.tar.gz|zip`, but the existing installer still downloads `https://erda-release.oss-cn-hangzhou.aliyuncs.com/cli/$OS/erda-$Version` as a raw executable and writes it directly to `/usr/local/bin/erda-cli` (`tools/cli/install.sh:55-58`). Once `make release-cli-publish` uses this publisher, the installer will either 404 or, if pointed at the new URL manually, install a compressed archive as the executable. Please update the installer to read the manifest/download/extract the archive, or keep publishing the legacy raw-binary paths during migration.

tools/release-cli/config.go
```diff
@@ -0,0 +1,101 @@
...
+	defaultArchList = []string{"amd64", "arm64"}
+	releaseTargets  = []releaseTarget{
```
luobily, line 32: The new release matrix only publishes `darwin/arm64` and `linux/amd64`. The current installer still treats all Darwin hosts as supported macOS without architecture detection (`tools/cli/install.sh:31-36`), so Intel Mac users will be silently dropped from the release path. If dropping `darwin/amd64` is intentional, please update the installer/docs and make the failure explicit; otherwise add a `darwin/amd64` target and build/publish it.

build/scripts/make-version.sh
```diff
@@ -11,25 +11,41 @@
 
 set -o errexit
 
+function normalize_version() {
+  local raw="${1/v/}"
```
luobily: `${1/v/}` removes the first `v` anywhere in the string, not only a leading `v`. For example `1.2-preview.1` becomes `1.2-preiew.1`, so a valid prerelease from `VERSION` or the `VERSION` env var can be silently corrupted. Please strip only a leading prefix, e.g. `${raw#v}` / `${raw#V}`, and add a small script test for prerelease strings containing `v`.

luobily (commented):
Reviewed the release tooling changes. I found a few release-chain compatibility issues that should be addressed before merging.

tools/release-cli/config.go
iutx: only support this list now.

luobily, line 32: The installer now detects both `x86_64|amd64` and `arm64|aarch64`, and maps Intel macOS to `darwin/amd64` (`tools/cli/install.sh`). But the publisher still only produces `darwin/arm64` and `linux/amd64`, so Intel macOS users will still go through the supported installer path and then fail to resolve any manifest for `cli/darwin/amd64/...`. Please either add a `darwin/amd64` release target/build output, or reject `darwin/amd64` explicitly in the installer with a clear unsupported-platform error.

luobily (commented):
Re-reviewed the latest update. The installer/archive path issue and version normalization issue look fixed; I left one remaining compatibility comment on the release matrix.

